PR #5, opened 2015-08-01
Make compatible with Python 3.3+

ariddell:
This might not work under Python 2.6 but it appears to be working under 2.7 and 3.4.

Thanks for the project. This is going to put an end to usb flash drives!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 language: python
 python:
   - "2.7"
+  - "3.3"
+  - "3.4"
 install:
   - pip install . Twisted pyflakes
 script:
--- a/src/wormhole/_version.py
+++ b/src/wormhole/_version.py
@@ -1,3 +1,4 @@
+from __future__ import print_function
 
 # This file helps to compute a version number in source trees obtained from
 # git-archive tarball (such as those provided by githubs download-from-tag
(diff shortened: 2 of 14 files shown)


warner:
Thanks! If we can figure out how to make this work with py2.6 too, I'll land it.

ariddell:
Might want to wait until `trial` gets ported to Python 3 -- which looks like it only happened 12 days ago, https://twistedmatrix.com/trac/ticket/5965

I'll look into testing on Python 2.6 soon.

warner:
Yeah, I don't think Twisted-on-py3 is complete enough for wormhole, even with `trial`. We need `twisted.web.client.Agent` and `twisted.web.server`, for the `EventSource`-based message-delivery layer, and according to https://twistedmatrix.com/trac/browser/trunk/twisted/python/dist3.py those aren't ported yet (`web/_newclient.py` *is* ported, which is probably >50% of the work, but unfortunately the main `Agent` is defined in `client.py`). I haven't looked at them closely, though, maybe they aren't too far off.

wormhole doesn't have much test coverage of the synchronous side (I couldn't figure out how to make the threads work right to allow the two ends to run at the same time). But we could at least use travis to test that there are no syntax errors against py3, and then do manual testing to make sure it actually works.

warner:
I just landed the py3-compatibility code today. Python 2.6/2.7/3.3/3.4/3.5 are all supported (and tested by Travis) for the non-Twisted blocking-style APIs. The Twisted APIs will have to wait until a few more parts of Twisted are ported to py3, specifically `twisted.web.client.Agent`. But I think that means we can close this one. Thanks!

ariddell:
Great. Glad to hear!

PR #14, opened 2016-06-02
add the tox  directory to gitignore

higs4281:


warner:
_trial_temp, good idea

PR #33, opened 2016-06-02
allow appveyor python3+ builds to fail

Laharah:
Added a small section to appveyor config file to allow the python3 windows builds to fail without warning of build failure. I'll look deeper into the windows issues to see if I can get the tests passing and then add the `[windows]` build extra. If I can I'll revert the config back.


warner:
thanks!

Issue #25, opened 2016-06-02
add dependency on twisted[windows_platform] when on windows, or on pypiwin32


warner:
Let's use the new PEP-508 syntax to declare a dependency on whatever windows-specific things we need (pypiwin32, I think) when we're building on windows.

meejah:
https://www.python.org/dev/peps/pep-0508/

warner:
That PR makes the windows instructions be `pip install magic-wormhole[windows]`. The next step is to use the PEP-508 syntax instead of the `[windows]` extra.

warner:
Fixed by a03c1804f7d4f6eff1150cb0ba4d730503c7d3b4 and PR #41. Thanks!

PR #35, opened 2016-06-02
added [windows] extra to encapsulate windows dependancies #25

Laharah:
adds the windows extra to the package that explicitly adds pypiwin32 as a dependency.

should be invoked with `pip install magic-wormhole[windows]` as per issue #25.


warner:
landed (in rebased form) in 13cda9e6159784dda7f7b67ae600fc64aac0fe0e . thanks!

PR #36, opened 2016-06-03
Prevent transmit from suggesting `127.0.0.1`

higs4281:
The test runs the listener locally, which can turn up 127.0.0.1.
The added code in transmit stops the hint.


warner:
Thanks! I've rebased and added a patch which allows 127.0.0.1 if that's the only item in the list. It looks like the Appveyor VMs *only* have 127.0.0.1, and the tests fail badly if there are no hints at all.

PR #40, opened 2016-06-03
zipfile perms

kneufeld:
when sending a directory the file permissions are not preserved. this is a deficiency with the python zipfile module so we need to work around it.


meejah:
LGTM, +1

warner:
Looks good. I added some more tests to exercise this, including a system-level test in `test_scripts`. Thanks!

PR #41, opened 2016-06-03
pep 508 and windows builds

Laharah:
This PR adds support for pep508 and removes the requirement for the `[windows]` extra to properly install on windows.

It also (finally) fixes the appveyor builds so that they run trial correctly. I'm not particularly happy about hard coding the `python src/wormhole/test/run_trial.py` call, but the behavior of -m is different between python 2 and 3 and this is the only way I could get it to work cross platform/version.

I can squash these commits for down for cleanliness if you want.


warner:
Looks good, I rebased to master and added a tweak to make the "test-coverage" environment do the same thing. THanks!

Issue #30, opened 2016-06-02
add `from __future__ import unicode_literals` to clean up source


warner:
The source tree is compatible with py2.6/3.3/3.4/3.5, which occasionally makes for some awkward constructs. There are a lot of strings declared as `u"foo"` instead of just `"foo"`. If we added `from __future__ import unicode_literals`, especially to `wormhole.py`, we could probably clean this up.

warner:
Ok, that's probably as much as we can convert in a single sitting: let's do other files (or new files) as we work on them for other fixes. Thanks @Laharah !

PR #44, opened 2016-06-04
Unicode literals pt2

Laharah:
combed out the `u"..."` literals from as many library modules as currently possible.

references issue #30


warner:
+1

PR #47, opened 2016-06-05
refactor to use Click

meejah:
Okay, this is ready now I believe. I left "wormhole-server" (e.g. instead of a subcommand) but we'd chatted about making it "wormhole server" so currently it's easy to go back (but if wormhole-server is the way to go then wormhole-server should get its own ``Config`` class)


warner:
- [ ] appease pyflakes
(travis, or rather tox, will complain until it's pyflakes clean)

cool stuff!

meejah:
Hmmm, here's a puzzler: when I run "tox -e py34", I get a failure for the ``test_version`` test, but if I run it by hand with ``./.tox/py34/bin/python -m wormhole.test.run_trial wormhole`` then it passes. Also tried after ``rm -rf .tox``. Let's see what Travis says ;)

meejah:
Hmm. I have no idea *why* that (last commit) works ... nor why other subprocess invocations apparently don't need it :/

meejah:
@warner Is this environment variable monkey-businesses what armin is getting at in all the Click (and blog-post) warnings against ``from __future__ import unicode_literals`` for py2/3 code?

(p.s. I presume the previous AppVeyor failure with \r\r\n vs \r\n is because the environment is extra-confusing for windows so I just compared after stripping whitespace)

meejah:
Okay, I just squashed everything of mine into one commit -- if you want to see the "original" branch I can push separately.

The only "thing to decide" for @warner is if we want to keep ``wormhole-server`` as a separate entry-point, or put it (back to) ``wormhole server``. If the former, it would be best to have a server-only ``Config`` class which I can put together pretty quickly (for now, it's re-using the "main" wormhole command's ``Config`` class).

warner:
Let's stick with the `wormhole-server` entrypoint for now.

warner:
I like it. Could you make (and perhaps rebase/rewrite) a few changes?:

* move the "remove unicode_data" changes into a separate patch (any idea why we didn't see this earlier? I remember static.Data being a part of Twisted that hadn't been fully ported to py3, but I also thought it'd been fixed by now.. what version of Twisted were you using when you saw this fail?)
* move the "throw no-contenders" exception in transit.py into a separate patch (or a separate PR, it's probably not too Click-related)
* cmd_server.py used to "return 1" when the current directory didn't have a twistd.pid . It'd be nice to still do `sys.exit(1)` in this case, so wrapper scripts can error out properly. I think that means raising a new exception after the EnvironmentError case, catching that exception from the click command handler, then.. hm, how do you control the exit code on a click handler anyways? Should it raise SystemExit?
* can we remove `argparse` from `setup.py` setup_requires= now?
* test_scripts got a line that says `XXX need captured stdin/stdout from sender/receiver`.. does the block of getvalues() just below not do that? (although for stdout+stderr, of course, there's so such thing as capturing stdin). If we leave an XXX line there, I want to make sure I understand what problem it's referring to
* I'm guessing we no longer have tests that make sure arguments are parsed correctly (test_scripts now starts with pre-parsed `Config` objects). Do you think we need tests like that? I haven't used Click enough to have a sense for it.
* RE the "XXX DeferredList" in test_scripts, I've found the syntax a bit easier to read when it does a pair of `yield x; yield y` statements, rather than `yield DeferredList(x,y,firstOnFirstError=True)`. Previously we *had* to use DeferredList because the oder of messages meant we weren't sure which side would fail first, and if we always waited on one of them (i.e. `yield x`), then sometimes we'd hang, since `y` would fail and `x` would wait until we closed y, or something. But with the new always-send-key-confirmation message that landed in 0.8.0, I think it's now safe to wait on just one thing at a time (which also exercises the order of failures, which is nice).
* `cli/cli.py _dispatch_command()`: typo, s/give/given/
* `cli/cli.py receive()`: uses `return` at end, but `send()` does not. I think `react()` doesn't return (it does os.exit when it's done), but python-mode (and the reader) don't know that, so maybe we should use `return react()` in both places, or just `react()` plus a comment to remind readers that it doesn't return.

meejah:
Yeah, I can move `unicode_literals` stuff. Not sure why it wasn't noticed in that refactor --  are there end-to-end server tests. Ditto no-contenders.

Yes, we can remove argparse.

XXX stuff is likely just leftover notes I didn't clean up.

I don't think we need "arg parsing" tests for code that isn't in magic-wormhole. In fact, it might be worth trying "prompt=True" for some of the prompting things, but I'm not sure that supports tab-completion.(Click can also take options from environment variables, e.g. could have WORMHOLE_RELAY).

warner:
Oh, that environment variable option could be nice. For local testing (offline, with a relay server running on my laptop), I've been using a shell script named `wormhole-local` that does `wormhole --relay-url=localhost/something "$@"`, which depends upon having `--relay-url=` live in a top-level option slot, rather than a sub-command slot. That makes `--help` less useful (`wormhole --help` shows it, but `wormhole send --help` does not), and would look pretty out-of-place if we ever add a subcommand that *doesn't* need the option (e.g. if switched from the distinct `wormhole-server` script back to a `wormhole server ..` subcommand).

warner:
I added some patches to fix some of those issues, check my 'pr47' branch and maybe pull it on top of yours.

Um, did you happen to do a manual test of the new parser? All the tests pass for me, but when I try a real `wormhole send FILENAME`, I get:

```
(ve) 12:warner@209-204-163-15% wormhole send docs/api.md
Sending text message (0 bytes)
Wormhole code is: 21-enrollment-spellbind

text message sent
```

Which suggests that it thinks it got `--text=""`, and it's ignoring the filename. The receiving side doesn't print anything, which is what I'd expect for an empty-body text-mode transfer.

Maybe a quick parser unit test is in order? And a change to the parser that distinguishes between `--text=""` and no `--text` argument at all?

meejah:
TODO:

- [ ] pull out unicode_literals fix
- [ ] send files
- [ ] add unittest for sending file
- [ ] "no contenders" in own PR
- [ ] exit status with no twistd.pid file
- [ ] test_scripts XXX comment
- [ ] return react()

meejah:
@warner what about c4cd4db for the twistd.pid thing?

meejah:
Okay, actually how about af209a5 for the error-stuff. Basically I changed `UsageError` into `InternalError` and used click.UsageError for "report this to the user nicely" exceptions.

warner:
Looks good, I'm adding a few patches from my pr47 branch to clean up some stuff (the port numbers in `wormhole-server start` were wrong, otherwise just cosmetics). Landing in  3.. 2.. 1..

PR #49, opened 2016-06-05
Issue29: notification on slow verification

### LINKED ISSUE #29: sender: if key-confirmation message doesn't arrive quickly, display pacifier message
There is a round-trip between the time the sender gets the receiver's PAKE message (and thus computes the session key), and the time the sender gets the receiver's VERSION message (and thus knows that the receiver used the right code). If this takes more than a second or two, it might be good to display a message like "Key established, waiting for confirmation..".

Laharah:
This branch should close #29 as it was described.

~~I tried to write a test for the scripts, but I couldn't monkey patch wormhole to slow it down without refactoring the script import structure completely. when manually testing I just put a sleep in the verification step.~~  Figured out a test after realizing I'm an idiot. I also put the message to stderr to keep piping clean.

I'm also not wild about the API verb `exchange_key` so let me know if you've got something better.

### DIFF
--- a/src/wormhole/cli/cmd_receive.py
+++ b/src/wormhole/cli/cmd_receive.py
@@ -69,7 +69,16 @@ class TwistedReceiver:
     @inlineCallbacks
     def _go(self, w):
         yield self._handle_code(w)
-        verifier = yield w.verify()
+        yield w.establish_key()
+        def on_slow_connection():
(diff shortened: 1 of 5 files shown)


Laharah:
rebased to upstream master.

warner:
Looks good.. sorry this took so long to land. I rebased to current master, and added a few small cleanup patches. Great work!

Issue #32, opened 2016-06-02
"ssh setup" feature


warner:
We came up with a neat feature idea at PyCon: using magic-wormhole to set up SSH pubkeys.

The use case is that Alice owns a computer, and wants to give Bob SSH acccess to it. Either Alice is root on the host and she's setting up a new account for Bob, or Alice is a normal user (logged in already) and is trying to add her own pubkey.

Alice runs something like `wormhole add-ssh`, maybe as `wormhole add-ssh --user=bob`. Then Bob runs `wormhole send-ssh`. The add-ssh command generates and displays a wormhole code. The send-ssh command looks in `~/.ssh/`, finds your pubkeys, and asks you which one you want to send, then accepts the wormhole code, and sends the pubkey. When add-ssh receives the pubkey, it appends it to `~/.ssh/authorized_keys` of the given user account.

kneufeld:
May I suggest `ssh-copy-id` as the command (or some subset of that) so it more closely matches the unix command of the same name. The whole looks like a duck thing... ;)

meejah:
Brainstorming about some basic use-cases. The simplest is:

A user with access to two machines wishes to enable machine A to ssh to machine B. So, they would:

1. run "wormhole ssh add" on machine B (yielding a wormhole code)
2. run "workhole ssh send" on machine A (transcribing the wormhole code)

Once the wormhole is established, `~/.ssh/identity.pub` is copied from machine A to machine B and appended to `~/.ssh/authorized_keys`.

jrollins:
+1 to @meejah's suggested implementation

warner:
We landed PR #57 just now. There are a few items I want to figure out before making a release with this feature:

* remove the verification prompt from the side that provides the SSH key
* add a verification prompt to the side that receives the SSH key (this is the side that is granting some authority, so they're the one that needs to be sure)
* the send/receive-a-blob utility functions that got added.. they currently take strings, rather than bytes. We should make that clear by adding a type check as a precondition assertion
* the current implementation defaults to adding the pubkey to the current user's `authorized_keys` file, creating it if necessary, and setting the permissions correctly to appease sshd. And there's an argument to point to a different username instead.
[...]

PR #51, opened 2016-06-07
idea, re: ssh command

meejah:
So I was thinking about the "wormhole ssh" command (#32) and it's really just a `wormhole send` + `wormhole receive` with some reading and writing of the ssh public-key and ~/.authorized_keys

...and so since you can invoke/forward commands in Click, I got a little carried away perhaps with this idea -- which just implements `wormhole ssh-add` by invoking `wormhole send` directly with a pre-chosen code, and `wormhole ssh-send` by invoking `wormhole receive` also directly via Click, with the public-key as a text-message.

Now, the current implementation here is a little awkward -- but maybe "send" + "receive" will be common-enough use-cases that they should just have a top-level Python API themselves...? (and then we can implement `wormhole ssh-add` directly using it).

That is, something like:

```python
# in package "wormhole"
def send(data, code=None, **kwargs):
    # guts of cmd_send.send() but without any input/output
    # returns None, or exception

def receive(code, **kwargs):
    # guts of cmd_receive.receive() basically, but no input/output
    # returns data, or exception
    # (or maybe it returns a dict with key data= and possibly others (like filename=, etc)?)
```


meejah:
(oh, note this is on top of the click-refactor branch so there's only 'really' one commit here)

warner:
Interesting. Yeah, as you noticed, the "nameplate id" (the numeric part at the start of the wormhole code") gets allocated by talking to the server, and there's a method in the Wormhole instance to help you out (which does both the nameplate allocation and the random-words selection).

A bunch of the options for send/receive would make sense for add-ssh/send-ssh too (--words, -0, --verify), but other than that, yeah, the ssh commands are basically pre/post wrappers around send/receive --text, so it feels like there ought to be a function they could call (which would also omit the file/directory processing code). I'm not sure what a good way to slice that file: making send-text into it's own function means duplication some code between text and file/directory.

Maybe there are some extra (hidden, non-Click-exposed) arguments we could pass into the send/receive functions, which could be used by wrappers? like.. a function that gets called at some point in the process? Maybe the prepare-file and write-file functions could be passed in as arguments, instead of being embedded in the handler? Dunno how clean that'd be.

meejah:
So I guess what I was getting at but didn't say very well: maybe "send a thing to the other party" is a common-enough use-case that it could be a top-level API thing in wormhole, that just returns a Deferred that fires with the data.

And then the `wormhole send` command itself as well as `wormhole ssh-send` would both use it under the hood. They would differ in that "the data" would always be an SSH public-key in the one case, but for the `wormhole send`, it would be either text, a file or a zip'd directory. Now, I'm not 100% sure this can be abstracted that far (or if it makes sense).

Another use-case for this API would be sharing Tahoe-LAFS introducer fURLs or MagicFolder invite-codes...

warner:
Ah, right, so like maybe a `OneShotWormhole` or a `WormholeTransaction`, which takes the most likely options as constructors (num-words, 0, verify, maybe some callables) and basically does call-and-response. I like it.

warner:
Oh, another thing occurred to me: the ssh commands should probably use a new "APPID", so that codes don't overlap with the "wormhole send/receive" commands (since `wormhole ssh-send` can't speak to `wormhole receive`). APPIDs are URL-shaped just for uniqueness management (there doesn't need to be any content at the URL.. maybe that makes them URIs? or URNs?). meejah, since you're implementing this, feel free to pick an APPID for it.

Also we should probably rebase the current patch on top of the Click rewrite that's already landed. I'm trying to do some click refactoring now, which moves the common arguments into a separate decorator (which would be applied to both `send` and `receive`). So maybe you should wait until that refactoring lands, so we can put the same decorator on add-ssh/send-ssh.

warner:
we can close this in favor of #57, right?

meejah:
yes

PR #54, opened 2016-06-18
unit-test for relay request, and fix Data() handling

meejah:
Pulled out of click-refactor. The reason this issue didn't come up with the unicode_literals changes is because there wasn't a test that just accessed the relay-server, so I added one for that too.


warner:
landed (in rebased form) in 07b4067727e97f182581965fe2ebc5a5ef172f75 . thanks!

PR #55, opened 2016-06-22
error if we have nothing to connect to

meejah:
pulled out of #47


warner:
Oops, I wound up fixing this slightly differently in the #47 branch. THanks!

PR #66, opened 2016-07-31
help bad spelers

dkg:
Some of us can never remember the old ditty:

   i before e, except after c
   or when sounding like "a"
   as in neighbor or weigh.

Perhaps magic wormhole can coddle us in our misorthography :)


warner:
Oh, that's awesome. Landing momentarily. Thanks!

PR #71, opened 2016-08-04
Use StreamServerEndpointService (available since Twisted 10.2)

dreid:
I was reading the code and I noticed this Service which is already built into twisted.


warner:
Oh! Thanks.. don't know how I missed that.

PR #78, opened 2016-08-15
Fix small CLI typo

flamsmark:
The output for verified copies asks the user to run `wormhole --verify receive`, but the correct argument order is `wormhole receive --verify`.


warner:
looks good.. will land in rebased form momentarily. Thanks!

Issue #91, opened 2016-11-09
do not attempt to download without required free space


anarcat:
i have had the bad experience of downloading 2.2 GB of data and then discovering i actually needed 2.3GB of free space on disk to download the whole file. the discrepency was due to an operator mistake in calculating (in my head) how much 2633026488 bytes were... (obviously, it would be nice to show human readable file sizes, but that's beside the point here).

wormhole shouldn't attempt to download a file that will fill the disk, because it will fail and it can't resume (#88).

warner:
Sounds reasonable to me, I'd accept a patch. I don't know how hard it is to do it right, though.. in Tahoe we did something similar, and it works horrendously differently on windows (see [allmydata.util.fileutil.get_disk_stats](https://github.com/tahoe-lafs/tahoe-lafs/blob/master/src/allmydata/util/fileutil.py#L443)). Alternatively we could just do `open(); seek(SIZE); write("\x00"); seek(0)`, but if the platform supports sparse files, that won't actually fail. We could do `write("\x00" * size)` (or a less memory-hungry equivalent), but that'd introduce a significant delay for large files. I can imagine having some highly asynchronous receiver running a loop that tries to populate the whole file (with zeros), and then when each new record arrives over the wire, it seeks back to the right place and writes the record, then seeks back to the end and keeps appending: that would avoid the delay, and would still fail earlier than before, but would still fail "late" sometimes, and the state machine would be kinda heinous.

The deeper problem is that we can't know the size of the incoming file until we've consumed the code, so the sender will always have to start again (and the receiver will always have to ask the sender to start again, which is a big crimp in the workflow). I have some ideas in mind for resumable transfers, but they'll depend upon using the same sending and receiving machines (it'll store a full-sized key in some small file on both of them).

anarcat:
On 2016-11-11 19:22:05, Brian Warner wrote:
[...]

PR #92, opened 2016-11-09
use human-readable sizes more broadly

anarcat:
there was a function to "abbreviate" sizes, but it was somewhat
unclear and incomplete. reuse the sizeof_fmt_* set of functions from
the borg backup project (MIT licensed) to implement a more complete
and flexible display that will scale up to the Yottabyte and
beyond. it also supports non-IEC units (like "kibibyte", AKA 1024
bytes) if you fancy that stuff.

this is a workaround for #91: it allows users to better see the size
of the file that will be transfered.

*some* places are still kept in bytes, most notably when receive fails
to receive all bytes ("got %d bytes, wanted %d") because we may want
more clarity there.

text transfers also use the "bytes" suffix (instead of "B") because it
will commonly not reach beyond the KiB range.

note that the test suite only covers decimal (non-IEC) prefix, but it
is assumed to be sufficient to be considered correct.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,6 +33,7 @@ setup(name="magic-wormhole",
           "autobahn[twisted] >= 0.14.1",
           "hkdf", "tqdm",
           "click",
+          "humanize",
       ],
       extras_require={
           ':sys_platform=="win32"': ["pypiwin32"],
--- a/src/wormhole/cli/cmd_receive.py
+++ b/src/wormhole/cli/cmd_receive.py
@@ -7,7 +7,7 @@ from twisted.python import log
 from ..wormhole import wormhole
 from ..transit import TransitReceiver
 from ..errors import TransferError, WormholeClosedError
-from ..util import dict_to_bytes, bytes_to_dict, bytes_to_hexstr
+from ..util import dict_to_bytes, bytes_to_dict, bytes_to_hexstr, sizeof_fmt_iec
 
 APPID = u"lothar.com/wormhole/text-or-file-xfer"
 
@@ -194,8 +194,8 @@ class TwistedReceiver:
                                                   file_data["filename"])
         self.xfersize = file_data["filesize"]
 
-        self._msg(u"Receiving file (%d bytes) into: %s" %
-                  (self.xfersize, os.path.basename(self.abs_destname)))
+        self._msg(u"Receiving file (%s) into: %s" %
+                  (sizeof_fmt_iec(self.xfersize), os.path.basename(self.abs_destname)))
         self._ask_permission()
         tmp_destname = self.abs_destname + ".tmp"
         return open(tmp_destname, "wb")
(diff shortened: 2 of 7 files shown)


anarcat:
i ran tests locally, they seem okay.

also note that one alternative would be to use [humanize](https://pypi.python.org/pypi/humanize) library, but it seems like a big overhead for something that is just a simple function. see also [this stackoverflow question](http://stackoverflow.com/a/1094933/1174784).

finally, note that some places may have been displaying "bytes" and will now display "B" instead.

meejah:
On the other hand, humanize is < 500 lines-of-code (not counting tests), so not really "big overhead".

anarcat:
true. it's just that it's yet another dependency. if you don't mind, i can just shove that in of course. it's just a search / replace away.

also note the [humanfriendly](https://humanfriendly.readthedocs.io/en/latest/) library which does similar things, but also allows for numbers input. that library is less widely distributed, unfortuantely: there's no debian package, for example...

let me know.

anarcat:
well shit, the tests don't pass. looks like there's some variation on the precision stuff in the different python version (not surprising because Python 3 changed the division semantics). Obviously, this fails with Python 2.7...

If you're okay with it, i'll reroll this with the humanize library instead.

warner:
Sounds like a good direction to go.

As long as it passes tests everywhere, I don't have a strong opinion about whether we use `humanize.filesize.naturalsize()` or not.. the format looks ok to me (I think `naturalsize(1234567) == "1.2 MB"`, where I'd probably go with `1.23 MB`, but I'm not going to lose any sleep over it), and it doesn't appear to pull in any sub-dependencies. `humanize` offers a lot of functionality we don't need, but it's not burdensome.

So I'd say do whatever's easiest. Including adding a `from __future__ import float_division` or whatever the magic is to make py2 behave like py3 (feel free to add such `__future__` lines in any module you like), if that helps resolve the py2-vs-py3 test failure in the code that you wrote.

anarcat:
the output format for `naturalsize` can be changed with the `format` option.

i pushed new code to use the humanize library, hopefully tests will pass without changes now.

warner:
that test failure looks pretty minor, a capitalization difference in the expected output

anarcat:
weird... i feel like i read that wrong; expected should not have a capital B and it doesn't, but whatever, i reversed capitalization, we'll see what travis thinks...

anarcat:
alas, tests seem to pass

warner:
Ok, looks good. I have a feeling that one of the `cmd_usage.py` changes is going to print the word "bytes" twice, but it's in a function that's been disabled anyways (uncorrected bitrot from several months ago), so I'm not going to worry about it. Landing momentarily. Thanks!

anarcat:
great, thanks!

PR #102, opened 2016-12-15
WIP: add manpages

### LINKED ISSUE #69: man pages
Most (all?) distros require man pages for binaries that are shipped with packages.  For Debian, we need man pages for wormhole and wormhole-server.  For the initial packaging that I've put together (https://bugs.debian.org/833090) I've made very basic man pages with help2man which do not include any subcommand help.  It would be great if we had a good way to generate man pages directly from the python built-in documentation that included subcommand usage for both wormhole and wormhole-server.

anarcat:
this PR adds manpages to the documentation. they are not hooked in the build yet - i am not sure how to ship those with setup.py, but it's a start and reduces the diff with debian.

i have improved the manpages a bit from what @jrollins had originally done for debian, by the way...

should close #69


warner:
Looks good, merging now. If I wanted to add `http://magic-wormhole.io` to the man pages somewhere, which section would be the best?

anarcat:
not sure - `SEE ALSO` maybe?

PR #109, opened 2016-12-17
send: flush stdout after displaying code

joeyh:
Fixes https://github.com/warner/magic-wormhole/issues/108

Apparently python defaults to buffering stdout when not outputting to
the terminal. It would be better IMHO if every line of stdout were
autoflushed, but based on
http://stackoverflow.com/questions/107705/disable-output-buffering
there's not a really good way to accomplish that in python.

### DIFF
--- a/src/wormhole/cli/cmd_send.py
+++ b/src/wormhole/cli/cmd_send.py
@@ -84,6 +84,8 @@ class Sender:
 
         if not args.zeromode:
             print(u"Wormhole code is: %s" % code, file=args.stdout)
+            # flush stdout so the code is displayed immediately
+            sys.stdout.flush()
         print(u"", file=args.stdout)
 
         yield w.establish_key()


warner:
Rats, even with `from __future__ import print_function`, python2.7 rejects `flush=True`. I want to keep this py2+py3-compatible, so that's out.

Happy to land this, although I think I'll change it to `args.stdout.flush()` to support the case where it's not writing to the real stdout. I'll have to test that first, make sure the `StringIO`s used in testing support flush.

meejah:
Setting `PYTHONUNBUFFERED=1` also causes Python to not buffer stdout (also, with `-u`). ...but I think you still have to do "something complicated" like re-opening stdout to get the currently-running Python to change its behavior:

```
sys.stdout = os.fdopen(sys.stdout.fileno(), 'w', 0)
```

(from http://stackoverflow.com/questions/107705/disable-output-buffering/1736047#1736047)

PR #116, opened 2016-12-23
transit: fix connect-to-self

warner:


warner:
that branch passes tests locally, but I need to do more testing before landing it, because it absolutely must continue to work for old clients (it's too easy to make a change that is self-consistent but not backwards compatible)

warner:
aha, and indeed it didn't work as expected. pushing a new branch momentarily

PR #120, opened 2016-12-26
Missing build dependency

xloem:


warner:
Thanks.. landing (in rebased form) momentarily. Any idea what the Fedora equivalent of `libssl-dev` might be? We should probably update the Fedora advice too, one of these days.

meejah:
It's probably "libssl-devel" but I'm not entirely sure. Paging @kneufeld ;)

kneufeld:
`openssl-devel`

warner:
@kneufeld thanks! What's the full command we should recommend.. `yum install python-pip build-essential python-devel libffi-devel openssl-devel`? I'm sure we need something different than `python-pip` and `build-essential` in there.

Also, where can I find a list of Fedora package names online? For debian and ubuntu I use https://www.debian.org/distrib/packages and http://packages.ubuntu.com/ . Is there some equivalent for the fedora world?

xloem:
Hi, I am installing on Fedora 23 right now.  Fedora does not have a `build-essential` packages, so I installed `gcc-c++` and `libtool` instead, which pull in cc, autoconf, automake, glibc, kernel headers, etc as dependencies.  I have found I also needed to install `redhat-rpm-config`.

On at least Debian I had to run `pip install --upgrade pip`, as the distribution provided version was too old to compile all the dependencies.  I have ran this on Fedora too, but I am unsure whether it is required.

On my Debian system behind a firewall designed to mostly only allow connections in and out via Tor, I was unable to get magic-wormhole to run successfully.  I would be able to provide shell access to this system for debugging to a trusted individual with extra time if desired.  But perhaps I just need to punch the right hole or set up the right hidden service.  It didn't "just work" with `torify wormhole`.

On Fedora, it is `dnf` now instead of `yum`.  So, a complete package installation command might be `dnf install python-pip python-devel libffi-devel openssl-devel gcc-c++ libtool redhat-rpm-config libsodium-devel` I believe.

I'm afraid I do not know where the fedora rpm lists are best reviewed online.  My local system downloads them from https://download.fedoraproject.org/pub/fedora/linux/releases/ .

kneufeld:
As of a release or two ago, `yum` is now `dnf`. No idea why they changed the name since as far as I can tell all the sub-commands and flags are the same. But I digress...

I usually just install `gcc` and `gcc-c++` but the Fedora equivalent of `apt install build-essential` is `dnf group install 'dnf group install 'C Development Tools and Libraries'`. I know right, rolls right of the figer tips. `dnf group list` for other goodies.  But I digress...

`dnf install python-pip etc` looks correct, just `build-essential` is different.

As for a package list I always just do a `dnf list > dnf.list` as the first command on a box. Faster to grep than `dnf search`. But I digress.

I haven't used it but you're probably looking for https://admin.fedoraproject.org/pkgdb/packages/

kneufeld:
Huh, this @xloem individual seems pretty bright. Especially since our two comments are virtually identical and posted 15 seconds apart. ;)

xloem:
Sorry for the accidental snipe!  What a coincidence.

kneufeld:
Again!

PR #129, opened 2017-01-04
Make footnotes section an HTML comment

toolness:
I noticed that the footnotes section appears as a visible heading when rendered as HTML (e.g. in GitHub), which looks odd because there's nothing below it, so I thought it might be useful to wrap it in an HTML comment so it's only seen by those who view it in its text form.  Feel free to close this PR if you disagree though!


warner:
looks good.. thanks!

PR #131, opened 2017-01-11
Add the packaging metadata to build the wormhole snap

elopio:
This is a package for the secure installation of apps that works in most Linux distributions.
Landing it upstream will enable builds for adventurous users.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1,0 +1,33 @@
+name: wormhole
+version: master
+summary: get things from one computer to another, safely
+description: |
+  This package provides a library and a command-line tool named wormhole,
+  which makes it possible to get short pieces of text (and arbitrary-sized
+  files and directories) from one computer to another. The two endpoints are
+  identified by using identical "wormhole codes": in general, the sending
+  machine generates and displays the code, which must then be typed into the
+  receiving machine.
+  The codes are short and human-pronounceable, using a phonetically-distinct
+  wordlist. The receiving side offers tab-completion on the codewords, so
+  usually only a few characters must be typed. Wormhole codes are single-use
+  and do not need to be memorized.
+
+grade: devel
+confinement: strict
+
+apps:
+  wormhole:
+    command: env LC_ALL=C.UTF-8 LANG=C.UTF-8 wormhole
+    plugs: [home, network, network-bind]
+
+parts:
+  magic-wormhole:
+    source: .
+    plugin: python
+    build-packages:
+      - gcc
+      - libffi-dev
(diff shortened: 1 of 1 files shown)


elopio:
You can find more information about snapcraft here: http://snapcraft.io/

To test it in an Ubuntu 16.04 machine:

    $ sudo apt install git snapcraft
    $ git clone https://github.com/elopio/magic-wormhole
    $ cd magic-wormhole
    $ git checkout snapcraft
    $ snapcraft
    $ sudo snap install *.snap --dangerous

If you have any questions or comments, please let me know.

warner:
looks good.. thanks!

elopio:
Thanks @warner.
Now you can register the name and maybe push an initial snap in the edge channel:
https://myapps.developer.ubuntu.com/dev/click-apps/register-name/

warner:
Does this lead to an Ubuntu package? Should we be leaving that task to an Ubuntu maintainer (like we do with Debian)? I've never used "snap" before, I don't know what the ecosystem looks like.

elopio:
@warner yes, that will get you an ubuntu package published in the store.

What we have learned after many years of packaging in ubuntu is that our 6 months/2 years release cycle and heavy-weight packaging requirements to make debs consistent, don't fit all the projects. So we are trying to remove the distro as a middle man, and let upstreams publish directly whenever they want. The distro maintainers are a safeguard in terms of security and api compatibilities in the archive, so, to make them unnecessary, snaps are totally isolated by default, and bundle all their dependencies. 

We are working hard to make the process as light-weight as possible so it won't be a burden to the developers. You can even hook the snap build and push to something like travis or launchpad to have it published automatically to the edge channel every time your repository changes. It will be a small effort one time, and then you can forget about it. Then when you want to make a release, you just run one command to move a revision from the edge channel to another one, on your own timeline.

So, magic-wormhole has made it to the zesty deb archive of ubuntu, and it will be in the next release we make in april. But most of our users are in xenial, the latest long-term support release, and they won't be able to get easily the magic-wormhole deb. With snaps, you push one version and it will work in xenial and any later updates. And we are working on porting it back and getting it into other distros, so it will work in ubuntu 14.04, debian, fedora, everywhere.

I'm sorry, I've just noticed I wrote a lot :) But I'm excited about this, and it would be great for us if you can give us some feedback, taking into account your experience with maintainers. We are in heavy active development, so that will help us improving the developer journey in Linux.

pura vida.

elopio:
ping @warner. Did you have a chance to push the snap to the store?

Issue #73, opened 2016-08-04
overwrite existing file when --output-file= is provided


warner:
A suggestion by `@teufen` on twitter: allow some way for the receiving user to allow an existing target file to be overwritten. This is especially useful if you're running the same `wormhole send` multiple times (with a newer version of the file).

It's not safe, in general, to overwrite existing files, especially because the filename is being selected by the sender. But we could give the receiver a way to opt-in to the overwrite, if it's happening predictably. And the existing `--output-file=` option is a pretty clear and predictable signal: any overwriting is going to affect a single user-specified path.

So the rule would be that if the recipient receives a file that already exists, but the filename came from `--output-file=`, then print a `overwriting FILENAME` message and overwrite it.

TBD: what to do when `--output-file=` is a directory: do we `rm -rf` the old directory and replace it entirely? (also, maybe this suggests that `--output-file=` is misnamed, if it could refer to directories)

horak:
So if I understand this thread correctly, `wormhole send --output-file <directory name>` is not at this point supported?

PR #139, opened 2017-02-05
rx: allow overwrite with --output-file= #73

smulloy21:
first time contributing to open source - let me know if anything's off :)

### DIFF
--- a/src/wormhole/cli/cmd_receive.py
+++ b/src/wormhole/cli/cmd_receive.py
@@ -257,8 +257,11 @@ class TwistedReceiver:
 
         # get confirmation from the user before writing to the local directory
         if os.path.exists(abs_destname):
-            self._msg(u"Error: refusing to overwrite existing '%s'" % destname)
-            raise TransferRejectedError()
+            if self.args.output_file: # overwrite is intentional
+                self._msg(u"Overwriting '%s'" % destname)
+            else:
+                self._msg(u"Error: refusing to overwrite existing '%s'" % destname)
+                raise TransferRejectedError()
         return abs_destname
 
     def _ask_permission(self):
--- a/src/wormhole/cli/cmd_receive.py
+++ b/src/wormhole/cli/cmd_receive.py
@@ -267,10 +267,8 @@ class TwistedReceiver:
         return abs_destname
 
     def _remove_existing(self, path):
-        if os.path.isfile(path):
-            os.remove(path)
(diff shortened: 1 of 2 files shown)


warner:
Wow, nice patch! And the tests are excellent.. great job!

One thought I had, maybe we should refactor `_decide_destname()` and `_ask_permission()` a bit, to make this cleaner. The case I'm thinking about is if you use `--output-file=`, and the file already exists, but you don't use `--accept-file`, and you say "no" when asked for permission. In that case, `wormhole receive` will delete the old file, but won't write anything into its place, which seem less than ideal (leaving you with less than you started).

I haven't figured out the best way to do this yet, but I'm wondering if we move the already-exists/overwrite checks into `_ask_permission()`, to make it easier to leave the file in place if permission is not granted. The tricky part is that the "Receiving file ..." message probably shouldn't print until we're really going to receive the file (so after permission is granted), but we need the filename and sizes to be displayed *before* asking permission.

Let me think about this. It might be easiest to just land your patch now, and then think about cleanups, especially since the denied-permission case isn't really that common.

thanks!

smulloy21:
@warner thanks!

for the denied-permission case, I have it only calling `_remove_existing()` from `_decide_destname()` if the `--accept-file` flag is present - otherwise it gets called after the permission check is successful in `_ask_permission()`. it's not the cleanest, cause it does the `os.path.exists()` check twice, but it should prevent the scenario you described - testing locally, when I use `--output-file=` but deny permission, the old file/dir isn't deleted

let me know if it works differently for you, or if you have ideas for a better way to handle it!

warner:
I did some manual testing.. looks like the old/new behavior is:


 --output-file | --accept-file | os.path.exists | old    | PR139
---------------|---------------|----------------|--------|--------
 no            | no            | no             | ask    | ask
 no            | no            | yes            | reject | reject
 no            | yes           | no             | write  | write
 no            | yes           | yes            | reject | reject
 yes           | no            | no             | ask    | ask
 yes           | no            | yes            | reject | **ask** (1)
 yes           | yes           | no             | write  | write
 yes           | yes           | yes            | reject | **write**

That looks great. And the "(1)" case I was wondering about does indeed leave the existing file alone if you say "no" when prompted.

In the future, we might simplify this a bit, by letting `--output-file=` imply `--accept-file`. But not today.

Landing momentarily. Thanks!

PR #141, opened 2017-03-18
typos

Sholes:


warner:
thanks! I wanted to somehow mark that "reactor" was an argument, but maybe some explicit markup would be better than that `=` sign. I'll add something on top of your patch.

PR #147, opened 2017-04-10
import from _wordlist which does exist

### LINKED ISSUE #146: wordlist vs _wordlist
There's a `_wordlist.py` in the repository and no `wordlist.py` afaict.

Importing `wormhole.codes` fails:

```
>>> import wormhole.codes
  File "<stdin>", line 1, in <module>
  File "/.../magic-wormhole/src/wormhole/codes.py", line 3, in <module>
    from .wordlist import (byte_to_even_word, byte_to_odd_word,
ImportError: No module named wordlist
>>>
```

exarkun:
Fixes #146

### DIFF
--- a/src/wormhole/codes.py
+++ b/src/wormhole/codes.py
@@ -1,7 +1,7 @@
 from __future__ import print_function, unicode_literals
 import os, six
-from .wordlist import (byte_to_even_word, byte_to_odd_word,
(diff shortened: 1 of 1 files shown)


warner:
`codes.py` should have been deleted, so I'm doing that instead of landing this. Good catch!

Issue #53, opened 2016-06-14
Rendezvous should not send list of in-use nameplates to clients


Laharah:
In a high volume relay, an attacker looking to intercept files can use the `_list` function to find nameplates waiting for a receiver, and use them to attempt file interception. On a high traffic relay, say 2-3 transfers a second, you have about a 30% chance of an intercept over the course of 2 hours. 

If we prevent auto-complete on nameplates however, an attacker would have to guess nameplates to find one with a sender listening. Moreover, if they are rapidly polling low nameplates, legitimate users would be less likely to receive those nameplates from the server, especially if the rate limiting is implemented on lonely nameplates.

warner:
Hm, I have to think about that. I find it really convenient to be able to tab-complete the nameplate id (the number at the start of the wormhole code), so it'd be sad to give that up. And I suspect that removing `list()` wouldn't slow down the rate at which the attacker could probe nameplates very much.

But yeah, if mass-polling of nameplates has the side-effect of preventing them from being allocated by subsequent clients, then that probably helps matters (at least it turns the MitM attack into a DoS attack, which is arguably better).

Removing tab-completion on nameplates would also simplify the code a bit, at the moment there's an awkward interface between the (blocking) readline-based tab-completion handler, and the (async) `list()` handler.

Laharah:
My one worry about rate-limiting is that it does make the relay even *more* susceptible to DOS attacks. It gets back to the problem of figuring out how many digits can a nameplate id have before it's inconvenient, forcing some kind of rollover on the oldest open nameplates. And then balancing that against how much we can rate-limit aborted nameplates before we make it too easy to DOS.

warner:
BTW I'm getting more comfortable with dropping tab-completion for nameplates.. they're pretty short. At the moment I usually type something like `4 TAB p u TAB s a TAB RETURN`, and this would change that into `4 - p u TAB s a TAB RETURN`. Not quite as pinky-bounces-on-TAB friendly (that hyphen is at the far other side of the keyboard), but not too bad.

#150 talks about making this configurable on the server, so some applications could choose to enable tab-completion of nameplates, and others could choose to turn it off.

Another thought I had was to add an attribute to each nameplate, saying whether it was willing to be included in the `list` (nameplates) output or not. Client-selected nameplates could opt out.

PR #151, opened 2017-05-08
Allow list option

meejah:
This adds a `--disallow-list` command-line option, with an "allow_list" in the code and a couple minor refactorings. See also #53

### DIFF
--- a/src/wormhole/server/cli.py
+++ b/src/wormhole/server/cli.py
@@ -48,9 +48,13 @@ def server(ctx): # this is the setuptools entrypoint for bin/wormhole-server
     "--signal-error", is_flag=True,
     help="force all clients to fail with a message",
 )
+@click.option(
+    "--disallow-list", is_flag=True,
+    help="never send list of allocated nameplates",
+)
 @click.pass_obj
 def start(cfg, signal_error, no_daemon, blur_usage, advertise_version,
-          transit, rendezvous):
+          transit, rendezvous, disallow_list):
     """
     Start a relay server
     """
@@ -61,6 +65,7 @@ def start(cfg, signal_error, no_daemon, blur_usage, advertise_version,
     cfg.transit = str(transit)
     cfg.rendezvous = str(rendezvous)
     cfg.signal_error = signal_error
+    cfg.allow_list = not disallow_list
 
     start_server(cfg)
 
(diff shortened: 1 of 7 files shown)


meejah:
Hmm, does the windows build just not work? Or did I somehow affect that util test with this patch??!!

warner:
I must have broken the windows build a few weeks ago, when I was trying to improve test coverage and did a `mock.patch()` on something that didn't exist on windows. If you rebase, it should be fixed now.

Could we put this functionality on `rendezvous.AppNamespace` instead of on the websocket-specific frontend server? I've got two motivations:

* future frontends (non-websocket) should honor this setting too
* some apps will want to allow this, some won't. (in fact really it's likely to be that some nameplates will want themselves to be tab-completable, and some won't, like offline/human-generated wormhole codes). So I'd like to move this closer to the nameplate record, and putting it in `AppNamespace` is as close as we can get without actually changing the DB schema yet.

Basically we want `AppNamespace.get_nameplate_ids` to return an empty list if nameplate-listing has been disabled. We need to rename the existing function to `_get_nameplate_ids` so it can continue to be used by `_find_available_nameplate_id` (which needs it to do allocation). Then the new `get_nameplate_ids()` will just be a wrapper that checks the permission flag and then either returns `self._get_nameplate_ids()` or `[]`. The permission flag should be passed down into the `AppNamespace` through `Rendezvous.get_app()`, so the `WebSocketRendezvousFactory` shouldn't see it at all.

If/when we make individual nameplates listable or not, I think we'll use an AND check. `_get_nameplate_ids()` will return an empty list if the app permission is turned off, else it will return just the nameplates that the DB record says are listable.

We've talked about allowing *allocation* to be enabled on a per-frontend basis, which would make some of these feature/permission switches want to live on a Factory, but I think I like our idea about API tokens better (so a co-located specially-blessed "client" uses its secret token to perform allocation, rather than using a special network socket that's unreachable from anywhere else).

Also, could you squash this down into one commit? Thanks!

meejah:
Okay, I updated as per the comment.

I also added some unit-tests.

warner:
looks good.. landing now

warner:
note to self: when reviewing, I missed the fact that the new `--disallow-list` option was only added to `wormhole-server start`, and not `wormhole-server restart`, causing today's server upgrade to have problems (quickly fixed by using stop+start instead of just "restart").

I should move the common arguments into a common class, and add test coverage for both `start` and `restart`.

PR #161, opened 2017-05-24
Add an "--ignore-unsendable-files" flag

### LINKED ISSUE #112: Attempting to send directory with broken symlinks causes error
Attempting to send a directory that contains broken symlinks makes `wormhole` choke.

Steps to reproduce (tested under Python 3.5.2 on Mac OS X 10.11.6):

```
$ python --version
Python 3.5.2 :: Continuum Analytics, Inc.

$ wormhole --version
magic-wormhole 0.8.2+27.gd81f2da

$ mkdir some_dir
$ ln -s nonexisting_file.txt some_dir/broken_symlink.txt
$ wormhole send some_dir
```

Resulting error message:

```python
Building zipfile..
  File "/Users/maxalbert/miniconda3/envs/wormhole/lib/python3.5/site-packages/twisted/internet/defer.py", line 651, in _runCallbacks
    current.result = callback(current.result, *args, **kw)
  File "/Users/maxalbert/miniconda3/envs/wormhole/lib/python3.5/site-packages/twisted/internet/defer.py", line 1355, in gotResult
    _inlineCallbacks(r, g, deferred)
  File "/Users/maxalbert/miniconda3/envs/wormhole/lib/python3.5/site-packages/twisted/internet/defer.py", line 1297, in _inlineCallbacks
    result = result.throwExceptionIntoGenerator(g)
  File "/Users/maxalbert/miniconda3/envs/wormhole/lib/python3.5/site-packages/twisted/python/failure.py", line 389, in throwExceptionIntoGenerator
    return g.throw(self.type, self.value, self.tb)
--- <exception caught here> ---
  File "/Users/maxalbert/work/code/other/magic-wormhole/src/wormhole/cli/cli.py", line 103, in _dispatch_command
    yield maybeDeferred(command)
  File "/Users/maxalbert/miniconda3/envs/wormhole/lib/python3.5/site-packages/twisted/internet/defer.py", line 1297, in _inlineCallbacks
    result = result.throwExceptionIntoGenerator(g)
  File "/Users/maxalbert/miniconda3/envs/wormhole/lib/python3.5/site-packages/twisted/python/failure.py", line 389, in throwExceptionIntoGenerator
    return g.throw(self.type, self.value, self.tb)
  File "/Users/maxalbert/code/magic-wormhole/src/wormhole/cli/cmd_send.py", line 55, in go
    yield d
builtins.FileNotFoundError: [Errno 2] No such file or directory: '/private/tmp/some_dir/broken_symlink.txt'
ERROR: [Errno 2] No such file or directory: '/private/tmp/some_dir/broken_symlink.txt'
```

toolness:
This is a **work-in-progress** attempt to fix #112 by:

* Making unreadable file errors more human-friendly;
* Adding an `--ignore-unsendable-files` flag that ignores such errors.

## Examples

Here's what it looks like if we send a directory called `breakage` with a broken symlink called `blap` in it:

```
$ wormhole send breakage
Building zipfile..
ERROR:  A file you wanted to send couldn't be read, maybe because it's
not a file, or because it's a symlink that points to something that
doesn't exist.  To ignore this kind of error, you can run wormhole
with the --ignore-unsendable-files flag.

blap: No such file or directory
```

And here's what happens when we use the new flag:

```
$ wormhole send breakage --ignore-unsendable-files
Building zipfile..
blap: No such file or directory (ignoring error)
Sending directory (22 Bytes compressed) named 'breakage'
Wormhole code is: 7-maritime-trojan
```

## To do

- [x] Add test to ensure `UnsendableFileError` is raised when `--ignore-unsendable-files` is not passed.
- [x] Add test to ensure `UnsendableFileError` is *not* raised when `--ignore-unsendable-files` is passed.

### DIFF
--- a/src/wormhole/cli/cli.py
+++ b/src/wormhole/cli/cli.py
@@ -10,7 +10,7 @@ from . import public_relay
 from .. import __version__
 from ..timing import DebugTiming
 from ..errors import (WrongPasswordError, WelcomeError, KeyFormatError,
-                      TransferError, NoTorError)
+                      TransferError, NoTorError, UnsendableFileError)
 from twisted.internet.defer import inlineCallbacks, maybeDeferred
 from twisted.python.failure import Failure
 from twisted.internet.task import react
@@ -109,7 +109,7 @@ def _dispatch_command(reactor, cfg, command):
         msg = fill("ERROR: " + dedent(e.__doc__))
         print(msg, file=cfg.stderr)
         raise SystemExit(1)
-    except WelcomeError as e:
+    except (WelcomeError, UnsendableFileError) as e:
         msg = fill("ERROR: " + dedent(e.__doc__))
         print(msg, file=cfg.stderr)
         print(six.u(""), file=cfg.stderr)
@@ -172,6 +172,10 @@ TorArgs = _compose(
 @click.option(
     "--text", default=None, metavar="MESSAGE",
     help="text message to send, instead of a file. Use '-' to read from stdin.",
+)
+@click.option(
+    "--ignore-unsendable-files", default=False, is_flag=True,
+    help="Don't raise an error if a file can't be read."
 )
 @click.argument("what", required=False)
 @click.pass_obj
--- a/src/wormhole/cli/cmd_send.py
+++ b/src/wormhole/cli/cmd_send.py
@@ -282,7 +282,8 @@ class Sender:
                         except OSError as e:
                             errmsg = u"{}: {}".format(fn, e.strerror)
                             if self._args.ignore_unsendable_files:
-                                print(u"{} (ignoring error)".format(errmsg))
+                                print(u"{} (ignoring error)".format(errmsg),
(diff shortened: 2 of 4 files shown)


toolness:
Ok, if travis/appveyor is happy with this, I think it's ready to go!

warner:
Looks like the windows error message is different:

 https://ci.appveyor.com/project/warner/magic-wormhole/build/1.0.733/job/y79ne9b5iiiyd74p#L464

I guess the test should assert that the error string is one of the two possible values.

toolness:
Oh shoot!  Ok, I think I fixed it, but I did so by making the test only test against the part of the string that doesn't include the actual OS error string--while testing for two would make things work across English versions of Linux and the version of Windows that appveyor uses, I believe it would fail on other locales and possibly other versions of OS's. Hope that's ok!

warner:
Looks good.. landing (in rebased form) momentarily.

PR #163, opened 2017-05-24
added two __main__.py for cli and server

kurtbrose:
I'm not sure if it is quite ready for merge yet, because the click help text is kind of messed up:

```
C:\Users\Kurt\workspace\magic-wormhole>python -m wormhole
Usage: __main__.py [OPTIONS] COMMAND [ARGS]...

  Create a Magic Wormhole and communicate through it.
```

That string should say `Usage: python -m wormhole [OPTIONS] COMMAND [ARGS]...`

Maybe I should file an issue with click? :-p

### DIFF
--- a/src/wormhole/__main__.py
+++ b/src/wormhole/__main__.py
@@ -1,0 +1,8 @@
+if __name__ != "__main__":
+    raise ImportError('this module should not be imported')
+
+
+from .cli import cli
+
+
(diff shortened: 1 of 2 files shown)


warner:
Heh. Naw, I think we can live with the confused synopsis.. let's fix that later. Merging momentarily. THanks!

PR #164, opened 2017-05-24
Update README.md

kennethreitz:
This is just a suggestion / my opinion, but I think this wording is much better, as it emphasizes the more impressive featureset of Magic Wormhole first. 

Hope you consider it! Many thanks for building this awesome tool. :)

:sparkles: :cake: :sparkles:


warner:
looks good.. thanks!

PR #165, opened 2017-05-25
Issue126 pow token

meejah:


meejah:
@warner is this heading in the right direction? It needs a bit of cleanup still and do we want to keep the token-database "not in-memory" even though this is essentially just placeholder for a "real" mitigation tactic?

PR #171, opened 2017-06-01
if the first connection fails, abandon the wormhole

### LINKED ISSUE #68: report initial connection failure and abort, rather than retrying forever
I'd expect an attempt to use a non-responsive or non-available relay to cause "wormhole send" to fail with an error code, but instead it hangs:

```
0 dkg@alice:~$ wormhole --relay-url ws://no-such-server/v1 send  --text 'bananas'
Sending text message (7 bytes)
Unhandled Error
Failure: twisted.internet.error.DNSLookupError: DNS lookup failed: Couldn't find the hostname 'no-such-server'.
```

I have to hit Ctrl-C to get out of this state.

warner:
This provides a clear error in case the user doesn't have an internet
connection at all, or something is so broken with their DNS or routing that
they can't reach the server. I think this is better than waiting and
retrying (silently) forever.

If the first connection succeeds, but is then lost, subsequent retries occur
without fanfare.

closes #68

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,7 @@ setup(name="magic-wormhole",
       install_requires=[
           "spake2==0.7", "pynacl",
           "six",
-          "twisted[tls]",
+          "twisted[tls] >= 17.5.0", # 17.5.0 adds failAfterFailures=
           "autobahn[twisted] >= 0.14.1",
           "automat",
           "hkdf",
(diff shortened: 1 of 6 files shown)


warner:
Note that this won't work until Twisted-17.5.0 is released, as it depends upon the new `clientService` API that it will provide.

warner:
Twisted-17.5.0 is out, this can land as soon as CI approves.

PR #181, opened 2017-06-26
Restart

warner:
fix `wormhole-server restart` w.r.t. the new --disallow-list argument


warner:
huh, not sure why this didn't automatically close.

PR #186, opened 2017-06-26
Add CLI path options

### LINKED ISSUE #184: Allow command-line configuration of the location of relay.sqlite and stats.json
It's operationally more complicated to have relay.sqlite, stats.json, and twistd.pid all in the same directory.

Let me do something like:

    $ wormhole-server start --relay-database-path /foo/bar.sqlite --stats-path /baz/stats.json

It wouldn't hurt to have an option for the location of the pid file but at worst that can always be controlled via working directory.

exarkun:
Fixes #184

### DIFF
--- a/src/wormhole/server/cli.py
+++ b/src/wormhole/server/cli.py
@@ -22,6 +22,15 @@ def server(ctx): # this is the setuptools entrypoint for bin/wormhole-server
     ctx.obj = Config()
 
 
+_relay_database_path = click.option(
+    "--relay-database-path", default="relay.sqlite", metavar="PATH",
+    help="location for the relay server state database",
+)
+_stats_json_path = click.option(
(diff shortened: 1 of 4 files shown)


exarkun:
This worked until I rebased onto 3b0a8de114f507e0b399d188a105f1ec5743fa42 (maybe bded01d3cce1fccb2fa037eb4698defd076e291b is the actual culprit).  I don't really know how click works or how this `Config` thing is meant to be used.  Why is a client `Config` being used for server stuff?  And why is it only broken now?  And groaaaan mocks `fake_start_server.mock_calls[0][1][0]`

warner:
It failed because `wormhole restart` needed the new arguments too, and I finally added a test to exercise that (which will, of course, be fixed by your second PR). I'll add a patch which adds the new args to `restart` too.

warner:
landing (in rebased form) momentarily

PR #187, opened 2017-06-26
Unify start and restart options

### LINKED ISSUE #185: `wormhole-server start` and `wormhole-server restart` options diverge
`start` accepts `--disallow-list` but `restart` does not.

Since `restart` proceeds to invoke the same codepath as `start` it seems these two should accept exactly the same arguments.  That they don't is a not-too-surprising consequence of all the options being defined twice, once for each command.

exarkun:
Fixes #185


exarkun:
Note this is stacked on #186

warner:
landing (in rebased form) momentarily

PR #189, opened 2017-06-27
Avoid corrupting state if creating a new db fails

### LINKED ISSUE #188: Creation and initialization of the relay SQLite3 database are not atomic
In database.py, `get_db`:

```
    must_create = (dbfile == ":memory:") or not os.path.exists(dbfile)
    try:
        db = sqlite3.connect(dbfile)
    except (EnvironmentError, sqlite3.OperationalError) as e:
        raise DBError("Unable to create/open db file %s: %s" % (dbfile, e))
    db.row_factory = dict_factory
    db.execute("PRAGMA foreign_keys = ON")
    problems = db.execute("PRAGMA foreign_key_check").fetchall()
    if problems:
        raise DBError("failed foreign key check: %s" % (problems,))

    if must_create:
        log.msg("populating new database with schema v%s" % target_version)
        schema = get_schema(target_version)
        db.executescript(schema)
        db.execute("INSERT INTO version (version) VALUES (?)",
                   (target_version,))
        db.commit()
```

If no database exists and the process crashes any time after `sqlite3.connect` but before `db.commit` then the system ends up in an inconsistent state from which it cannot recover without manual intervention - a database file exists but has no schema.  Subsequent attempts to open it will fail, most likely with an error about the `version` table not existing.

For example:

```
starting wormhole relay server
  File "/app/env/bin/wormhole-server", line 11, in <module>
    sys.exit(server())
  File "/app/env/local/lib/python2.7/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/app/env/local/lib/python2.7/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/app/env/local/lib/python2.7/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/app/env/local/lib/python2.7/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/app/env/local/lib/python2.7/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/app/env/local/lib/python2.7/site-packages/click/decorators.py", line 27, in new_func
    return f(get_current_context().obj, *args, **kwargs)
  File "/app/env/local/lib/python2.7/site-packages/wormhole/server/cli.py", line 74, in start
    start_server(cfg)
  File "/app/env/local/lib/python2.7/site-packages/wormhole/server/cmd_server.py", line 42, in start_server
    twistd.runApp(c)
  File "/app/env/local/lib/python2.7/site-packages/twisted/scripts/twistd.py", line 25, in runApp
    _SomeApplicationRunner(config).run()
  File "/app/env/local/lib/python2.7/site-packages/twisted/application/app.py", line 380, in run
    self.application = self.createOrGetApplication()
  File "/app/env/local/lib/python2.7/site-packages/twisted/application/app.py", line 440, in createOrGetApplication
    ser = plg.makeService(self.config.subOptions)
  File "/app/env/local/lib/python2.7/site-packages/wormhole/server/cmd_server.py", line 24, in makeService
    allow_list=self.args.allow_list,
  File "/app/env/local/lib/python2.7/site-packages/wormhole/server/server.py", line 43, in __init__
    db = get_db(db_url)
  File "/app/env/local/lib/python2.7/site-packages/wormhole/server/database.py", line 57, in get_db
    raise DBError("db file is unusable: %s" % e)
wormhole.server.database.DBError: db file is unusable: no such table: version
```

(Yes, actually encountered this in real usage.)

exarkun:
Fixes #188 

The approach here is to initialize new databases at a temporary path and then atomically rename them when that has succeeded.

I can think of other possible approaches:

  * delete the database file if an exception is encountered.  this strikes me as dangerous (ie, having a painful  consequence if the logic is wrong and the wrong thing is deleted).
  * re-arrange the logic more significantly so that an empty database is treated just like a non-existing database file is currently treated: as needing to have its schema initialized.  I could probably go for this (it's how Axiom works, iirc, and I don't remember any problems with that).

But the approach in the linked branch also seems to work.

### DIFF
--- a/src/wormhole/server/database.py
+++ b/src/wormhole/server/database.py
@@ -1,6 +1,7 @@
 from __future__ import unicode_literals
 import os
 import sqlite3
+import tempfile
 from pkg_resources import resource_string
 from twisted.python import log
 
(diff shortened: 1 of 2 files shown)


warner:
Looks good to me.. I'll rebase to current master and land.

PR #200, opened 2017-06-28
Make default option of prompt uppercase

lindskogen:
Makes it clear what happens when user just presses `Enter`

See: https://stackoverflow.com/questions/7803728/standard-format-for-yes-no-questions-in-the-terminal


warner:
oh, good idea

PR #202, opened 2017-06-28
Use the latest git tag as the snap version

elopio:
With this change it's easier to automate the continuous delivery of the snap, because when a verified tag is pushed to the repo the snap will get that version. There will be no need to manually change the version in the snapcraft.yaml.


warner:
cool, thanks. Landing (in rebased form) momentarily.

PR #206, opened 2017-06-29
improve installation instructions

kryptogeist:
Added separated subsections for each OS (OS X, Linux, Windows) on the installation guide.


warner:
That looks awesome.. thanks!

Landing (in rebased form.. the commit history was getting too messy) momentarily.

PR #219, opened 2017-07-10
import contextlib, random

cclauss:
Also on line 239, where is `parse()` defined and on line 241, where is `update_my_state()` defined?


warner:
ah, good catch.. I'll land this. Thanks!

demo-journal.py is still an experiment: it doesn't actually run yet. (it's more of a sketch to help me visualize what the eventual API wants to look like). `parse()` and `update_my_state()` are placeholders for what would be application-specific behavior.

I'm hoping to swing back around to "journaled mode" in a few months, and at that point I'll try to make this example more functional.

PR #220, opened 2017-07-10
Add flake8 testing only to Python 3.6

cclauss:


warner:
Let's point this at just `src/` and `setup.py`: running it on `.` will get all of the `.tox/` virtualenvs too, which will take forever and not add anythng useful. Travis was complaining because the files in `misc/` weren't py3-safe, but I'm inclined to skip everything in misc/ for now and maybe add them later after the real code is flake8-clean.

Also, how about we just add it to tox, instead of travis? It will be easier to fix the warnings that it discovers if we can run the same test locally. How about we use `tox -e flake8` for now?

cclauss:
@warner Can you please and accept this PR as is before I open a separate PR to attack the tox piece?  I find tox tricky to configure correctly but if you are a tox master then I would be happy to learn from your implementation of this piece.  In the end, flake8 should be run under both Py2 and Py3 as it will return different results under the two different versions.

warner:
The changes look good.. could you do two things first?:

* squash this down into two patches: one for misc/demo-journal.py, the other for `.travis.yml`
* rebase to current master

The appveyor failure should be fixed on master (it was a windows-specific problem that I introduced several weeks ago, and fixed yesterday).

Then force-push to this same "patch-2" branch that you're using, and travis+appveyor will pick it up. Once CI is green, I'll land it. And then after it's landed, we can figure out how to extend it to run under both py2 and py3, hopefully via tox somehow. Thanks!

cclauss:
Closed in favor of #229

PR #225, opened 2017-07-21
Added Other Requirement for Installing on Windows.

Buckaroo9:
Microsoft Visual C++ 9.0 is required for installation. Otherwise large errors are show when executing pip install magic-wormhole.


warner:
Cool, thanks, landing in rebased form momentarily.

PR #228, opened 2017-07-23
flake8 tests in .travis.yml

cclauss:


warner:
Weird.. for some reason, the appveyor windows py3 tox is not installing all the dependencies that it ought to (it's missing autobahn). I can't think of any reason why changing `.travis.yml` could possibly do that. I'm going to restart the appveyor build and see if anything changes.

warner:
I'm seeing the same problem at home. Apparently something is broken in tox, or pypi, or setuptools.. not the fault of this patch.

warner:
Looks like setuptools-36.2.1 was released about an hour ago (in the middle of the Appveyor run, which is why py27 passed and all the subsequent py3 builds failed). I've filed https://github.com/pypa/setuptools/issues/1099 about it, sounds like they pushed out an emergency fix for some other dependency-tracking issue and this was a side-effect.

Let's wait a day or two for them to fix it, and then re-trigger these builds.

cclauss:
Understood... #220 is OK?

warner:
Not quite, I see that you removed the travis changes from #220, but could you also squash it and rebase to current master? You'll need to force-push it to your own branch once you've done that. There should be exactly one patch on that branch, and it should only touch the demo-journal.py file. (basically we don't need a lot of noisy "changed this, oops changed it back", etc, in the branch that gets merged, just a single easy-to-read patch)

PR #237, opened 2017-07-30
use versioneer to set the version of the snap

### LINKED ISSUE #235: snap packaging version should come from Versioneer
The `snapcraft.yaml` packaging control file has a line that says `version: git`, which is good because it means I don't need to manually update a version string inside `snapcraft.yaml` each time I do a release. We use Versioneer on the setup.py, which means I don't have to update a version string there either: I just `git tag` and push.

It'd be nice if the snapcraft version always matched the setup.py version. They're computed in very similar ways (they both use some flavor of `git describe`), but in a perfect world they'd always be identical.

@elopio: is there any way to tell `snapcraft.yaml` to run a script to learn what version it should use? It could just use the output of `setup.py --version`.

elopio:
closes #235.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1,5 +1,6 @@
 name: wormhole
 version: git
+version-script: python3 -c "import versioneer; print(versioneer.get_version())"
 summary: get things from one computer to another, safely
 description: |
   This package provides a library and a command-line tool named wormhole,


warner:
Thanks for the quick response! Could we have it to `python setup.py --version` instead of using Versioneer directly? That should be future-proof against changes to setup.py that use something other than versioneer.

elopio:
The problem is that setup.py requires setuptools. versioneer on the other hand doesn't require any package that are not included in python3.
This `version-script` executes in the context of the build, so by the time we call the version-script, setuptools is not installed.

We have discussed about a way to set the version at different times during the build cycle, not just at the start. But we don't want to add new features just because they sound cool, before they are really needed by our users. 

It would be great f you can explain your use case in forum.snapcraft.io. That will help us assign the right priority for this new feature.
You might find it interesting the discussion that started here:
https://forum.snapcraft.io/t/snapcraft-version-scripts/106

warner:
Ah, ok. I guess I was assuming that setuptools is included in the python2/3 package, but of course it's still a separate package.

I'll land this, and I'll check out that forum and explain how Versioneer works. Thanks!

PR #256, opened 2017-10-04
Relay and transit url from env vars

jtdoepke:
Added the click `envvar` kwarg to look for relay and transit helper urls in environment variables. I'm running my own relay/transit servers inside a corporate firewall. This would make life easier since I can just add the urls to my bashrc and forget about them.


warner:
Looks great.. and an excellent test too. Thanks!

PR #267, opened 2017-12-14
Change the send command output to help copy and pasting to others

### LINKED ISSUE #266: Improve information message when sending
Hi,

Just a thought, but it would be nice to have the command and the code to send on one line - this makes copy and pasting easier when sending the information to a remote party. 

So, from:

`On the other computer, please run: wormhole receive`
`Wormhole code is: 1-foo-bar`

to

`On the other computer, please run: wormhole receive 1-foo-bar`

This is easier to copy and paste and send :) 

-=david=-

dharrigan:
A tiny update to show the command to execute along with the code required on
the same line, rather than split across two lines. This small change helps
when sending the information to others using copy and paste.

Fixes #266

-=david=-

### DIFF
--- a/src/wormhole/cli/cmd_send.py
+++ b/src/wormhole/cli/cmd_send.py
@@ -100,9 +100,6 @@ class Sender:
             args.code = u"0-"
             other_cmd += " -0"
 
-        print(u"On the other computer, please run: %s" % other_cmd,
-              file=args.stderr)
-
         if args.code:
             w.set_code(args.code)
         else:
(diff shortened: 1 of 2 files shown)


warner:
Sounds like a good direction. I think I'd like to retain the bit that basically says "this is a wormhole code", since that reinforces the docs and the mental model of the code being the important thing to transfer (when we get a real GUI application, the recipient won't be typing "wormhole receive", they'll be double-clicking an application icon and then typing the code into some text field).

What about making the output look like:

```
Sending 123kB file named foo.txt
Wormhole code is: 1-revival-rematch
On the other computer, please run:

wormhole receive 1-revival-rematch

(output waits here until connection is made)
```

So the command to cut/paste is on a line of its own, separated by blank lines on both sides.

Also, please update the test to match: look in src/wormhole/test/test_cli.py, there are assertions that look at the output of wormhole-send and make sure it matches expectations. Those need to be updated to match the new output.

thanks!

dharrigan:
Hi,

I like your suggested text improvements :-) Code updated. If you could cast your eye over the tests please - they all pass now. Not too sure about the last one, the text one.

-=david=-

warner:
Looks good.. I made some tweaks to simplify the code flow and improve the `--text` and `-0`-mode output too. Landing momentarily. Thanks!

PR #271, opened 2018-01-01
Do not refer to transit server in Dockerfile

jml:
https://github.com/warner/magic-wormhole/pull/263 removed the transit server,
breaking the Dockerfile.

This fixes the Dockerfile to run just the rendezvous server.


warner:
good catch, will apply.

PR #274, opened 2018-01-21
Make ipaddress dependency conditional on python2

hlieberman:
Note, right now this patch will be an no-op because txtorcon has not
yet released a new version which contains this patch.  Once it has
done so, however, this will make the build dependency exclude
ipaddress on python3 environments.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,6 +25,7 @@ setup(name="magic-wormhole",
               "wormhole-server = wormhole.server.cli:server",
           ]
       },
+      setup_requires="setuptools>=36.2",
       install_requires=[
           "spake2==0.7", "pynacl",
           "six",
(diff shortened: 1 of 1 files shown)


hlieberman:
FYI @anarcat

warner:
Could you look at the appveyor failure on py3.4 (https://ci.appveyor.com/project/warner/magic-wormhole/build/1.0.952/job/0vkrqmwxtdfgipju at the bottom of the log) and see if there's anything you can do about that?

warner:
Do we know why `ipaddress` was needed in the first place? I see it showed up when we added tor, and originally it was only included in the `extras_require= [tor]` feature. Then I did some refactoring (2d949f23) that (accidentally?) made it an unconditional requirement. Is this something that txtorcon needs? Should txtorcon be expressing this requirement internally? It doesn't look like there's any code in magic-wormhole that imports `ipaddress`.

warner:
Local unit tests pass if I just remove `ipaddress` from `setup.py` entirely. I haven't tried actually using the `--tor` support, though.

meejah:
txtorcon does express this requirement: https://github.com/meejah/txtorcon/blob/master/requirements.txt#L5

PR #277, opened 2018-01-25
Add snapcraft support

popey:
This pull request adds support to build magic-wormhole as a snap.

magic-wormhole is already [available](https://snapcraft.io/wormhole/) in the snap store via the yaml in the [snapcrafter](https://github.com/snapcrafters/magic-wormhole) repo.

This PR is to upstream the snapcraft.yaml which creates the snap which is in the store. If the PR is accepted, we can transfer ownership of the magic-wormhole snap to you, the upstream developer. Once done you could use the free build.snapcraft.io service to automatically build new releases of magic-wormhole and push directly to the store. 

If accepted, I'm happy to walk you through the next steps.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,32 @@
+name: wormhole
+version: git
+summary: get things from one computer to another, safely
+description: |
+  This package provides a library and a command-line tool named wormhole, which
+  makes it possible to get arbitrary-sized files and directories (or short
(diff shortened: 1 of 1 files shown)


warner:
Hey, sorry I didn't dive into this earlier. It looks like there's currently a `snapcraft.yaml` in the root directory (not in `snap/`).. is that good neough? Are there any changes we should make to it in addition to (or instead of) this PR?

abitrolly:
@popey your reply is needed, sir.

PR #279, opened 2018-02-08
appveyor.yml: try different artifact syntax

### LINKED ISSUE #278: appveyor "artifact" has broken name (binary wheels)
Appveyor is a windows CI service (like Travis, but only for windows). We're using it for tests and to build binary wheels, since I don't have a windows machine of my own. There's a section in the `.appveyor.yml` config file that's supposed to upload the generated `.whl` file into the Appveyor "Artifacts" archive, but the config is doing the wrong thing, since the generated filename has an embedded backslash: `dist\magic_wormhole-0.10.4-py2.py3-none-any.whl`.

https://ci.appveyor.com/project/warner/magic-wormhole/build/job/uy5a5vmu59huqqav/artifacts

I don't yet know what the correct syntax is, to get a variably-named file in the `dist/` directory. On Tahoe-LAFS we tried just `- path: dist`, and that produces a single zipfile named `dist.zip` which contains the whole directory. What I want is to get separate artifact files for every file inside `dist/`, and I don't know the name of the files ahead of time (since they'll depend upon the version number and commit id).

warner:
will close #278

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -51,7 +51,7 @@ after_test:
 
 artifacts:
   # bdist_wheel puts your built wheel in the dist directory
-  - path: dist\*
+  - path: 'dist\*'
 
 #on_success:
 #  You can use this step to upload your artifacts to a public website.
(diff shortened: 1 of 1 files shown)


warner:
nevermind

Issue #287, opened 2018-02-23
report client version to mailbox server


warner:
I've avoided doing this so far, but I think I need to have the client report its version to the mailbox/rendezvous server so I can measure new-release uptake. I don't currently have any visibility at all into what versions people are running, and I expect it'll help me plan out deprecation of old API versions if I knew what percentage of clients were capable of speaking the new versions.

My plan is to add a "version" key to the BIND message (which is the first message that clients send to the server upon each websocket connection). The value will be a 2-tuple of (implementation, version), with `("python", "0.10.5")` being the current value. The server will record (side, implementation, version, rounded_time) in the usage database, and tooling will de-duplicate `side` before aggregating the results (to avoid double-counting clients which lose their websocket connection and have to make a new one). (`side` is a random string, mostly used to avoid talking to ourselves, but also for deciding which side is the "leader" and which is the "follower" in the upcoming Dilation protocol).

Baeumla:
I'm working on this right now. 

Do you think it is enough to just send `("python", "0.10.5")` or should we also include the python-version the client is using with `sys.version` instead of only the implementation `"python"`?

warner:
Thanks for grabbing this one!

I'd just do `("python", "0.10.5")`. Let's add a table for developers somewhere that says`"python"` means `github.com/warner/magic-wormhole` (or wherever else it ends up living), and if someone starts an independent python implementation then we can call that one `python2` or `python-$authorname` or whatever (i.e. the word "python" is more of a pointer to the particular implementation, rather than implying that the language it's implemented in is particularly important).

Use `wormhole.__version__` for the second string, that'll get updated by Versioneer automatically.

We'll need code on the mailbox/rendezvous-server side to add the two versions (one for each side of a successful connection) into the Usage database. That'll be tracked in https://github.com/warner/magic-wormhole-mailbox-server/issues/3 . But the server will ignore any unrecognized keys in API messages, so we can add this to the client completely independently of it being actually used by the server.

I'm now on the fence about whether to send a 2-tuple or to merge the strings together first (`python-0.10.5`), but I guess it's better to send the 2-tuple and let the server decide how it's going to record the data.

warner:
Done. Thanks!

PR #293, opened 2018-03-08
report client version on bind

Baeumla:
ref #287

I had to create a little workaround for the `versions` dictionary in the wormhole `create` to get the version from there. I did so since I tried to avoid imports at `_rendevouz.py` or `_boss.py` and to be less invasive. 

Please let me know if you had any other ideas about this or if there are ways to improve the code.

The message the server recieves looks like this:
`Bind message: {'appid': 'app_id', 'implementation': 'python', 'wormhole_version': '0.10.2+127.g86edf01.dirty', 'type': 'bind', 'id': 'c7cc', 'side': '41d8cb019d'}`

### DIFF
--- a/src/wormhole/_boss.py
+++ b/src/wormhole/_boss.py
@@ -32,12 +32,11 @@ class Boss(object):
     _url = attrib(validator=instance_of(type(u"")))
     _appid = attrib(validator=instance_of(type(u"")))
     _versions = attrib(validator=instance_of(dict))
-    _wormhole_version = attrib(validator=instance_of(type(u"")))
+    _client_version = attrib(validator=instance_of(tuple))
     _reactor = attrib()
     _journal = attrib(validator=provides(_interfaces.IJournal))
     _tor = attrib(validator=optional(provides(_interfaces.ITorManager)))
     _timing = attrib(validator=provides(_interfaces.ITiming))
-    _implementation = attrib(validator=instance_of(type(u"")))
     m = MethodicalMachine()
     set_trace = getattr(m, "_setTrace", lambda self, f: None) # pragma: no cover
 
@@ -55,8 +54,7 @@ class Boss(object):
         self._RC = RendezvousConnector(self._url, self._appid, self._side,
                                        self._reactor, self._journal,
                                        self._tor, self._timing,
-                                       self._wormhole_version,
-                                       self._implementation)
+                                       self._client_version)
         self._L = Lister(self._timing)
         self._A = Allocator(self._timing)
         self._I = Input(self._timing)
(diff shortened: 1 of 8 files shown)


warner:
Thanks for diving into this! A couple of thoughts:

There are three separate kinds of versions here (I apologize for not
naming them very well, I wasn't thinking of the future when I wrote that
code):

* "Application Versions", which are passed into the `create()`
  constructor via the `versions=` argument, come from the application,
  and are delivered to the application on the other end through the
  `w.get_versions()` API. This is a way for the two applications to
  negotiate their shared abilities.
* "Wormhole Versions", which are created by the wormhole library, sent
  through the encrypted channel in a phase=VERSION message, and are
  delivered to the wormhole library at the other end. They are not
  delivered to the application, but rather they're for the two wormholes
  to negotiate internal things like the upcoming "dilation" feature. The
  Application Versions are stored inside the Wormhole Versions
  dictionary, in a key named `app_versions`.
* the new "Client Version", which will be created by the wormhole
  library, included in the BIND message (not encrypted, at least not by
  the shared PAKE key), and delivered to the Mailbox Server. This is not
  seen by the remote side at all: it's consumed by the server.

We don't need to have the application pass in the Client Version; it
should be computed internally. So:

* let's leave cmd_send and cmd_receive alone, so no new arguments being
  passed into `create()`
* in `create()`, let's leave the `wormhole_versions` value alone
  (dictionary with a single `app_versions` key copied from the
  `versions=` argument)
* let's name the new field `client_version`, and make it a tuple of
  `(implementation, version)`, rather than two separate keys. So the
  current release would be `("python", "0.10.5")`
* let's use `from . import __version__` to pull in the version directly
* I admit that I'm not sure if Versioneer (which creates `_version.py`)
  provides bytes or strings, especially on py2-vs-3. You're absolutely
  right to prepare for either. But let's not switch on Python2-vs-3,
  let's just say that if `__version__` is bytes, treat it as UTF-8 and
  decode it (but tolerate problems, so use `errors="replace"`), and if
  it's str/unicode, leave it alone. That should work for both py2 and
  py3. Feel free to use something from the `six` module if that helps.
* pass the new string (always unicode at this point) into Boss with a
  new `client_version=` argument, instead of `wormhole_version`, and
  from the Boss into the RendezvousConnector
* deliver the new string in the BIND message as a single new argument
  named `client_version=` (all the keyword arguments get turned into
  JSON keys)
* let's leave `test_wormhole.py` alone: if applications want to pass
  non-strings into their `versions=` argument, that's their prerogative,
  the wormhole library should just deliver the value unmodified to the
  other application
* remove the extra `b. start()` change in `wormhole.py`

good work.. thanks so much for taking this one on!

Baeumla:
Okay, thank you for your feedback on this :) . I'm working on it

warner:
Looks good.. pushing (with a few cosmetic changes, like removing the debug print in the tests) in a moment. Thanks!

PR #296, opened 2018-04-21
Pep8 conformance - A Reference PR

copyninja:
Most files are made to converted using YAPF (Yet Another Python Formatter) and isort is used to sort the imports and remove unused imports if any. Currently there is E741 which is ignored using tox.ini as its sort of not felt okay to fix the variable name from l to something else as l and 1 looks similar in some fonts.

Currently build will fail because flake8 has McCabe complexity detection enabled. There are some part of code which seems to be too complex as complexity is higher than 10. (a standard default used).

This PR is probably just for reference and might not get merged.

cc: @warner


ThomasWaldmann:
Typo: *conformance*

Suggestion: don't do PRs that fail builds (that mccabe complexity), but either disable that or configure it to a level so builds will pass and open a ticket with a TODO to enable / lower that again (including fixing the src so it won't fail then either).

copyninja:
@ThomasWaldmann I've clearly mentioned the intentions of PR and also mentioned it might break. Since @warner  wanted to see how it looks I created PR. 

So yes I know it breaks and its clearly mentioned in description and thanks for your suggestions.

ThomasWaldmann:
@copyninja I didn't say anything else, I was just suggesting a potential way to have a more merge-likely PR (IMHO).

copyninja:
@ThomasWaldmann yes I understand. But this PR has some context, this is originated from discussion with @warner on IRC on PEP8 confirmance of code and some related tests. He wanted to see how it looks. And failing error is purposely left because it is highlighting some complex piece of code if I remove the test then it will be hidden.

And no when I say not meant to be merged it is actual intention. That is why I mentioned `A Reference`

warner:
No worries.. I've been poking at this today, I'm going to land a version of it that fixes a few more things. I don't entirely agree with pep8, but I'm ok with targetting it for uniformity.

`autopep8` doesn't seem to fix everything that `flake8` complains about: I think I've been spoiled with how well `rustfmt` works over in Rust land.

And flake8 presents some impossible situations like:

```python
if ("foo" in stuff and
    "bar" not in stuff):
    do_things()
```

(where the conditional is too big to fit on a single line)

flake8 complains that the `do_things()` visually lines up with the inside of the conditional statement, which is true, but it also doesn't accept some extra whitespace in the second half of the condition since then the second half won't line up with the first half. The only way I've found to appease it is to use *two* parentheses, which just seems silly.

Also for files like `test_ipaddrs.py`, I'd like to retain the original (overlength) lines of sample output from `ifconfig` rather than wrap them with `\` to address the long-line complaint. But there's no way to disable the long-ine complaint for just a single line inside a literal block-quoted string, nor is there a way to disable just one warning on a per-file basis. So I've left the line-wrapping in place for now.


Anyways, I'm going to land this, and see how it feels going forward.

PR #298, opened 2018-06-02
use pyinstaller to build wormhole fat/standalone binary

ThomasWaldmann:
tested with/on:
- ubuntu linux 18.04 amd64
- pyinstaller 3.3.1 (pip install pyinstaller)
- python 3.6.5

There is a good chance it also works on FreeBSD, maybe also on macOS.

The change in `__main__.py` was required because otherwise it complains about `__main__` not being a package when trying the dot-relative import.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,12 +24,6 @@ var/
 MANIFEST
 .eggs
 
-# PyInstaller
-#  Usually these files are written by a python script from a template
-#  before PyInstaller builds the exe, so as to inject date/other infos into it.
-*.manifest
-*.spec
-
 # Installer logs
 pip-log.txt
 pip-delete-this-directory.txt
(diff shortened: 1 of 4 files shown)


ThomasWaldmann:
Would be cool if some FreeBSD and macOS users could test this.

PyInstaller also supports Windows, but adapting for Windows might be a bit more work, not sure.

warner:
Cool, thanks!

ThomasWaldmann:
For borg, we use vagrant to build platform binaries, in case you are interested you could grab it from there and build and release fat binaries for convenient download and usage.

See there:

https://github.com/borgbackup/borg/blob/1.1-maint/Vagrantfile

I sign releases so users can verify the binaries are as they were built, so a release looks like this:

https://github.com/borgbackup/borg/releases/tag/1.1.6

There are no windows binaries as borg does not support native windows, but pyinstaller supports building native windows binaries.

PR #300, opened 2018-06-02
update Windows dependency

ofek:
`pywin32` moved to GitHub and now ships binary wheels itself.

https://github.com/mhammond/pywin32
https://pypi.org/project/pywin32/#files


warner:
good catch, thanks for the update

PR #304, opened 2018-06-12
Add long_description_content_type to setup.py

### LINKED ISSUE #303: PyPI page for magic-wormhole is empty
Currently https://pypi.org/project/magic-wormhole/ looks like this:

ekrano nuotrauka is 2018-06-12 14-04-23

Would you please consider passing the contents of README.md as `long_description` in setup.py (and set `long_description_content_type='text/markdown'`, so it'll be rendered correctly)?

mgedmin:
The new PyPI code discards long_description when it's not valid ReStructuredText.  Luckily it also supports Markdown, but you have to [specify the content type](https://packaging.python.org/tutorials/packaging-projects/#creating-setup-py) explicitly in that case.

(IIRC for this to work correctly you also have to use sufficiently recent versions of setuptools, wheel, and twine to upload releases: see https://github.com/pypa/python-packaging-user-guide/issues/481.)

Should fix #303.


warner:
ah, I was wondering why so many projects have a blank description. Thanks for the fix.

mgedmin:
Oh no, this didn't work!  :(

ekrano nuotrauka is 2018-11-12 15-38-13

warner:
Hrm. Could you point me at another project that has this working correctly? I'll compare setup.py and formatting and see if I can figure out what's going wrong.

mgedmin:
https://pypi.org/project/requests/ uses Markdown (https://github.com/requests/requests/blob/master/setup.py#L74)

They also recently had problems with it, and it turned out to be an issue with an older version of twine used to do the upload: https://github.com/requests/requests/issues/4859#issuecomment-435929879

Issue #311, opened 2018-09-23
`send --text -` reads in all the text before sending


joshtriplett:
I wanted to stream over a pipe, so I tried piping into `wormhole send --text -`, but that seems to read in all the data first, so it ran out of memory.

warner:
Yeah, sorry, the send-text protocol currently only sends a single message, and doesn't support streams.

When we get the "Dilation" protocol (#312) implemented, then streams should be pretty easy, and `--text -` won't need to buffer everything first.

PR #323, opened 2019-02-22
Added support for sending block devices

WeirdCarrotMonster:
One day i needed to dump 8GB SD card and send it to colleague's PC. Since piping data doesn't work yet (#311) and python doesn't handle block devices as files, i added a check for st_mode.


warner:
Huh. Yeah, ok, why not.. thanks!

PR #331, opened 2019-05-18
Closes #330 Changed default choice from No to Yes for wormhole receive

### LINKED ISSUE #330: switch confirm-receive prompt default to "yes"?
As @morrieinmaas mentioned in #327, it's a bit surprising that, in addition to typing "wormhole receive" and the code, you *also* have to say "yes" (or "y") at the confirmation prompt.

I made it default to "no" because I figured this is the last chance to protect yourself against an unexpectedly huge download, or with a filename that's going to be surprising (some dot file with scary side effects). But really, it's unlikely that you're going to say no after coming all this way.

So I think maybe in the next release we could switch this to a default "yes". Users still have a chance to bail out once they see the filename/size, but they'd have to type more to escape at that point.

morrieinmaas:
I've simply changed the default for receive to Yes. Unless specifically typing `n` or `N` receive defaults to accept receive upon carriage return.

### DIFF
--- a/src/wormhole/cli/cmd_receive.py
+++ b/src/wormhole/cli/cmd_receive.py
@@ -369,8 +369,8 @@ class Receiver:
     def _ask_permission(self):
         with self.args.timing.add("permission", waiting="user") as t:
             while True and not self.args.accept_file:
-                ok = six.moves.input("ok? (y/N): ")
-                if ok.lower().startswith("y"):
+                ok = six.moves.input("ok? (Y/n): ")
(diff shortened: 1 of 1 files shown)


warner:
Looks good, thanks! I've tweaked it a bit so that typing "y" or just return accepts the transfer, but typing anything else cancels it, so if someone panics and types "quit" or "help" or something like that, it will cancel too.

morrieinmaas:
Sounds good - thanks for adding the tweak.

PR #332, opened 2019-06-04
Fedora has magic-wormhole in all versions

hroncok:


warner:
awesome, thanks for the update

PR #361, opened 2019-09-29
Add installation instructions for openSUSE

aspiers:
magic-wormhole has been included in openSUSE since Leap 15.1, and is
also available in Tumbleweed:

https://software.opensuse.org/package/python-magic-wormhole

So document this explicitly.


warner:
thanks!

PR #378, opened 2020-02-06
Reactor import improvements

sneakypete81:
This PR removes a module-level import of twisted.internet.reactor. This import has side effects, which can cause problems when using a non-default reactor (qt5reactor, for example).

For me this led to a bug where transit would connect via the relay but never directly to the peer. Changing the order of imports in the calling module fixed the problem, but this PR should stop it happening again.

### DIFF
--- a/src/wormhole/transit.py
+++ b/src/wormhole/transit.py
@@ -10,8 +10,9 @@ from collections import deque
 
 import six
 from nacl.secret import SecretBox
+import twisted.internet
 from twisted.internet import (address, defer, endpoints, error, interfaces,
-                              protocol, reactor, task)
+                              protocol, task)
 from twisted.internet.defer import inlineCallbacks, returnValue
 from twisted.protocols import policies
 from twisted.python import log
@@ -559,7 +560,7 @@ class Common:
                  transit_relay,
                  no_listen=False,
                  tor=None,
-                 reactor=reactor,
+                 reactor=None,
                  timing=None):
         self._side = bytes_to_hexstr(os.urandom(8))  # unicode
         if transit_relay:
@@ -579,6 +580,8 @@ class Common:
         self._waiting_for_transit_key = []
         self._listener = None
         self._winner = None
+        if reactor is None:
+            reactor = twisted.internet.reactor
         self._reactor = reactor
         self._timing = timing or DebugTiming()
         self._timing.add("transit")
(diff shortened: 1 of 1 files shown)


sneakypete81:
I tried the following in `_build_listener`:
```
-        ep = endpoints.serverFromString(reactor, "tcp:%d" % portnum)
+        ep = endpoints.serverFromString(self._reactor, "tcp:%d" % portnum)
```

But this led to the test failure below, so I changed the above back to use the default reactor instead
of the one passed into the constructor. 

```
[ERROR]
  File "/home/travis/build/warner/magic-wormhole/src/wormhole/test/test_transit.py", line 1318, in test_success_direct
    hints = yield s.get_connection_hints()  # start the listener
  File "/home/travis/build/warner/magic-wormhole/.tox/py37-coverage/lib/python3.7/site-packages/twisted/internet/defer.py", line 1418, in _inlineCallbacks
    result = g.send(result)
  File "/home/travis/build/warner/magic-wormhole/src/wormhole/transit.py", line 618, in get_connection_hints
    direct_hints = yield self._get_direct_hints()
  File "/home/travis/build/warner/magic-wormhole/src/wormhole/transit.py", line 660, in _get_direct_hints
    d = self._listener.listen(f)
  File "/home/travis/build/warner/magic-wormhole/.tox/py37-coverage/lib/python3.7/site-packages/twisted/internet/endpoints.py", line 491, in listen
    return defer.execute(self._reactor.listenTCP,
builtins.AttributeError: 'Clock' object has no attribute 'listenTCP'

wormhole.test.test_transit.Transit.test_success_direct
```

src/wormhole/transit.py
```diff
@@ -579,6 +580,8 @@ def __init__(self,
         self._waiting_for_transit_key = []
         self._listener = None
         self._winner = None
+        if reactor is None:
+            reactor = twisted.internet.reactor
```
warner: This will probably fail.. if `twisted.internet` had a property named `reactor`, it would be loaded already and we'd lose the benefit of delayed loading. Maybe remove the `import twisted.internet` from the top, and replace this with `from twisted.internet import reactor`

src/wormhole/transit.py
```diff
@@ -596,7 +599,7 @@ def _build_listener(self):
         direct_hints = [
             DirectTCPV1Hint(six.u(addr), portnum, 0.0) for addr in addresses
         ]
-        ep = endpoints.serverFromString(reactor, "tcp:%d" % portnum)
+        ep = endpoints.serverFromString(twisted.internet.reactor, "tcp:%d" % portnum)
```
warner: Yeah, you're right, this should really be `self._reactor`. I'll see if I can figure out what was making that test fail.

src/wormhole/transit.py
sneakypete81: You're right - the code above only works because twisted sneakily removes the twisted.intenet.reactor module and replaces it with an attribute of the same name. I prefer your suggestion as it doesn't rely on this madness!

PR #384, opened 2020-04-12
Replace deprecated attrs `cmp=False` parameter with `eq=False`

sneakypete81:
See https://attrs.readthedocs.io/en/stable/api.html#core

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -49,7 +49,7 @@ setup(name="magic-wormhole",
       install_requires=[
           "spake2==0.8", "pynacl",
           "six",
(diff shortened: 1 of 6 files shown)


warner:
Could you also bump the `setup.py` dependency on `attrs` to 19.2.0? According to their Changelog, that's the first version which offers the new `eq=` parameter.

It looks like CI failed because `pyflakes` got better at spotting `.format` errors, and there were two pre-existing failures there.. I've landed your PR for it, thanks. If you rebase this PR on new trunk, we should get a clean CI run.

sneakypete81:
Good catch - done. Feel free to update the `setup.py` comment as you see fit though.

PR #394, opened 2020-09-18
Fix link to rendezvous server repo in API doc

ClaytonPassmore:
Fixes an issue where the link to the rendezvous server repository was being rendered as plain text in the docs.

I formatted the link weirdly to keep the line length down :shrug:


warner:
cool, thanks!

Issue #403, opened 2021-01-26
Update development command for zsh users


adam-sroka:
Hi,

I recently tried running magic-wormhole on macOS using the (very nicely documented) development instructions in [welcome.md](https://github.com/magic-wormhole/magic-wormhole/blob/master/docs/welcome.md). However, running `pip install -e .[dev]` results in an error, since zsh uses [square brackets for globbing / pattern matching](http://zsh.sourceforge.net/Guide/zshguide05.html#l137). The same issue has been discussed [here](https://github.com/mu-editor/mu/issues/852) and can be fixed by running `pip install -e .'[dev]'` instead.

Since latest macOS versions ship with zsh as default, it might be useful to add a note about this to [welcome.md](https://github.com/magic-wormhole/magic-wormhole/blob/master/docs/welcome.md), so that other people who are trying to develop magic-wormhole on a Mac don't run into this issue.

Please see my PR that addresses this and let me know what you think 😊 .

PR #404, opened 2021-01-26
Update development pip command for zsh

adam-sroka:
This PR fixes Issue #403.


meejah (approved):
:+1: LGTM
(I don't have merge-permission or I would have done that too)

thanks for the fixes!

Issue #395, opened 2020-09-21
Misleading "You could try again" after failed "wormhole receive"


jwilk:
This is what happens when you misspell your code for `wormhole receive`:
```console
$ wormhole receive 6-exodus-unwinf
ERROR:  Key confirmation failed. Either you or your correspondent
typed the code wrong, or a would-be man-in-the-middle attacker guessed
incorrectly. You could try again, giving both your correspondent and
the attacker another chance.
```
The "You could try again" part, if you don't read it carefully, or if you don't understand well how MW works, sounds like an invitation to correct your typo:

```console
$ wormhole receive 6-exodus-unwind
```
Not only this will not work, but (thanks to nameplate reuse — #31) it could also break somebody else's rendezvous.

warner:
Good point. What if we enhanced that to say "try again from the beginning (with a different code)" ? Or "try sending the file again"?

sagarhowal:
I added the "try sending the file again". Do you want to add the "attacker might have a chance to guess password again" in the message as well?
PR #410

PR #410, opened 2021-04-07
update WrongPasswordError

sagarhowal:
changed error message to say "Try to send file again" as discussed in issue #395 .
Not sure if the caution message about "your attacker may have another change to guess" should be added. It makes the message too long.


meejah (approved):
LGTM

PR #411, opened 2021-04-20
add a --debug-state command-line option

meejah:
This exposes set_trace functionality for debugging
state-machine transitions in the various state machines.

### DIFF
--- a/src/wormhole/cli/cli.py
+++ b/src/wormhole/cli/cli.py
@@ -38,6 +38,30 @@ class Config(object):
         self.stdout = stdout
         self.stderr = stderr
         self.tor = False  # XXX?
+        self._debug_state = None
+
+    @property
+    def debug_state(self):
+        return self._debug_state
+
+    @debug_state.setter
+    def debug_state(self, debug_state):
+        if not debug_state:
+            return
+        valid_machines = [
+            'B', 'N', 'M', 'S', 'O', 'K', 'SK', 'R', 'RC', 'L', 'C', 'T'
+        ]
+        debug_state = debug_state.split(",")
+        invalid_machines = [
+            machine
+            for machine in debug_state
+            if machine not in valid_machines
+        ]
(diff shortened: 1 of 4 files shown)


piegamesde:
**BIKESHED** Why don't you call the flag "verbose" or something more common? There's no need to restrict trace level messages to those that somehow inform about state changes.

meejah:
`set_trace` is an Automat method; these aren't "trace"-level log messages, they're specifically for tracing the state-transitions of the various state-machines (so are very unlikely to be useful to non-developers .. is why I hid it behind a thing starting with `--debug` ...)

piegamesde (changes requested):
So I gave it a quick try and noticed the following:

- If I don't type a state name, the error message isn't any helpful (but maybe that's restricted to the CLI argument parsing framework)
- The value of the option is called `TEXT`. There is no indication of what it means and what the possible values are.
- I just checked, and I can give that option an arbitrary value. So maybe make it optional with a sane default?
- You probably have renamed the option, the first commit shows an old name.

meejah:
Great, thanks for taking a look .. I'll make some fix-ups (and yeah, on a second look it's not very apparent what one should pass to it ... I guess a good default would be "trace ALL the machines")

meejah:
okay, made some fixups and re-worked it a little.

meejah:
I re-worked this a bunch .. if you're interested in a second look, that'd be great @piegamesde

piegamesde:
Sorry I can't test right now, but from the code documentation the change looks sane. I'd argue that using the full state machine names instead of one/two letter codes would be better, but I think I can see why you are doing this from a technical standpoint.

meejah:
yeah .. honestly I don't really expect anyone to go beyond `--debug-state` (or at least: I haven't).
In general I agree about the names, but even in the code they're all referred to by those same one or two letters (`self._B.<whatever>` for example)...

PR #418, opened 2021-05-24
welcome.md: hyperlink changes

ProactiveServices:
Replaced 404ing pages with alternatives, upgraded http: to https:, replaced redirected with direct links, added missing hyperlinks, moved some in to context. The Microsoft VC page infers that direct links to VC++ 9.0 may move/die later, so perhaps best to link to the general download page? Virtualenv link is 404 so I've replaced it with what I presume is an equivalent.


warner:
Looks great, thanks!

PR #422, opened 2021-07-03
Send empty directories in `wormhole send` too

### LINKED ISSUE #232: Empty directories
Wormhole does not send empty sub-directories (I attempted from Windows to MacOSX).

CosmicHorrorDev:
Closes #232 

Hello! Obligatory thanks for the great tool, it makes sending directories between computers work like a dream

I ended up running #232 which confused me for a bit since I wasn't expecting empty directories to be silently excluded.

This change ended up touching more than I was expecting, but reusing the logic for adding files or directories to the zipfile required pulling some things out to separate functions.

In regards to styling I  couldn't find any documentation listing style guidelines or formatters used, so I just kept changes to an 80 character line limit (and it looks like `cmd_send.py` accidentally got auto-formatted without me realizing it).

The changes do pass the test suite for me (python 3.9 on linux), and beyond that everything seemed to work fine when testing manually. I didn't test any other versions since I don't have any other python interpreters installed and this is my first time dealing with Python 2 code, so any subtle issues there may have slipped by me.

I'd also like to note that this does change the message on the uncompressed size since it's now including directories too which are typically 4KiB each. I think this would be fine since the message is talking about the uncompressed size, but if you want it to just be for the size of files then I'm fine with changing things around to better conform to how things used to be.

### DIFF
--- a/src/wormhole/cli/cmd_send.py
+++ b/src/wormhole/cli/cmd_send.py
@@ -252,10 +252,12 @@ class Sender:
         archive_path = os.path.join(*tuple(local_path + [entry_name]))
         local_entry_path = os.path.join(base_path, entry_name)
 
-        num_bytes = 0
+        entry_bytes = 0
+        entry_added = False
         try:
             zip_file.write(local_entry_path, archive_path)
-            num_bytes += os.stat(local_entry_path).st_size
+            entry_bytes += os.stat(local_entry_path).st_size
+            entry_added = True
         except OSError as e:
             errmsg = u"{}: {}".format(entry_name, e.strerror)
             if self._args.ignore_unsendable_files:
@@ -265,7 +267,7 @@ class Sender:
             else:
                 raise UnsendableFileError(errmsg)
 
-        return num_bytes
+        return entry_bytes, entry_added
 
     def _zip_dir_offer(self, what, basename):
         print(u"Building zipfile..", file=self._args.stderr)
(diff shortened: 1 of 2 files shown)


LovecraftianHorror:
Looks like there is a failure with broken symlinks. I'll fix it tomorrow, not sure why I didn't see it when running the tests before

LovecraftianHorror:
It looks like the issue was actually that ignored files would still be added to `num_files`. The simple fix was to just have `_add_zip_entry()` also return a bool that indicates if the entry was added and to check the bool when incrementing `num_files`

LovecraftianHorror:
Huh, weird that appveyor said that it succeeded when the 3.7 job failed. Looks like the build environment needs an update

From: https://ci.appveyor.com/project/warner/magic-wormhole/builds/39857754

```text
generating cffi module 'build\\temp.win-amd64-3.7\\Release\\_openssl.c'
    running build_rust
    error: can't find Rust compiler
    
    If you are using an outdated pip version, it is possible a prebuilt wheel is available for this package but pip is not able to install from it. Installing from the wheel would avoid the need for a Rust compiler.
    
    To update pip, run:
    
        pip install --upgrade pip
    
    and then retry package installation.
    
    If you did intend to build this package from source, try installing a Rust compiler from your system package manager and ensure it is on the PATH during installation. Alternatively, rustup (available at https://rustup.rs) is the recommended way to download and update the Rust compiler toolchain.
    
    This package requires Rust >=1.41.0.
    
        =============================DEBUG ASSISTANCE=============================
        If you are seeing a compilation error please try the following steps to
        successfully install cryptography:
        1) Upgrade to the latest pip and try again. This will fix errors for most
           users. See: https://pip.pypa.io/en/stable/installing/#upgrading-pip
        2) Read https://cryptography.io/en/latest/installation.html for specific
           instructions for your platform.
        3) Check our frequently asked questions for more information:
           https://cryptography.io/en/latest/faq.html
        4) Ensure you have a recent Rust toolchain installed:
           https://cryptography.io/en/latest/installation.html#rust
        5) If you are experiencing issues with Rust for *this release only* you may
           set the environment variable `CRYPTOGRAPHY_DONT_BUILD_RUST=1`.
        =============================DEBUG ASSISTANCE=============================
```

meejah:
Thanks for the PR!
I haven't reviewed it, but that Windows thing is familiar and almost certainly _not_ the fault of your code...I've clicked the button to let the other CI run.

LovecraftianHorror:
It looks like the only "failure" was a slight reduction in code coverage

I'd be happy to add a test for sending a directory that includes sending an empty dir if you can point out a test that I could base it off of. From a very brief look I couldn't find an integration test to base it off of, but I think that would be a good fit for testing the behavior

Adrian999999999 (approved):

PR #423, opened 2021-07-22
Permissions implementation

meejah:
WIP: this is a proof-of-concept implementation of the "permissions" spec, implementing the "none" and "hashcash" methods.

Needs:

- [ ] unit-tests
- [ ] cleanup

### DIFF
--- a/src/wormhole/_boss.py
+++ b/src/wormhole/_boss.py
@@ -223,6 +223,9 @@ class Boss(object):
     #   something we said badly, or due to CrowdedError)
     # * "error" is when an exception happened while it tried to deliver
     #   something else
+    # * "permission-required" is for DoS etc mitigation and allows the
+    #   server to ask for additional information / work to proceed. We
+    #   only support "hashcash" or "none" currently.
     def rx_welcome(self, welcome):
         try:
             if "error" in welcome:
@@ -234,6 +237,12 @@ class Boss(object):
             # middle of processing the rx_welcome input, and I wasn't sure
             # Automat would handle that correctly.
             self._W.got_welcome(welcome)  # TODO: let this raise WelcomeError?
+
+            permission = welcome.get("permission-required", {})
+            if "hashcash" in permission:
+                self._RC._send_hashcash(permission["hashcash"])
+            self._RC._send_bind()
+
         except WelcomeError as welcome_error:
             self.rx_unwelcome(welcome_error)
 
--- a/src/wormhole/_rendezvous.py
+++ b/src/wormhole/_rendezvous.py
@@ -28,6 +28,7 @@ class WSClient(websocket.WebSocketClientProtocol):
 
     def onMessage(self, payload, isBinary):
         assert not isBinary
+        print("onMessage {}".format(payload))
         try:
             self._RC.ws_message(payload)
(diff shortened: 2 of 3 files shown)


vu3rdd (approved):
Looks great.

src/wormhole/_rendezvous.py
```diff
@@ -61,6 +62,54 @@ def buildProtocol(self, addr):
         return proto
 
 
+def mint_hashcash(bits, resource):
+    "1:6:210623:arbitrary string::9WrCxB1SdCBOM3i5:000005"
```
piegamesde, line 66: What is this line for?

src/wormhole/_rendezvous.py
```diff
@@ -61,6 +62,54 @@ def buildProtocol(self, addr):
...
+    counter = 0
+    timestamp = "210623"
```
piegamesde, line 68: Since the server provides the `resource` part in a challenge-response design, checking the time stamp is not actually required for this to work. Nevertheless, I'd suggest you to include a current time anyways because the server side might use an off-the-shelf library that checks it regardless.

src/wormhole/_boss.py
```diff
@@ -234,6 +237,12 @@ def rx_welcome(self, welcome):
             # middle of processing the rx_welcome input, and I wasn't sure
             # Automat would handle that correctly.
             self._W.got_welcome(welcome)  # TODO: let this raise WelcomeError?
+
+            permission = welcome.get("permission-required", {})
+            if "hashcash" in permission:
```
piegamesde, line 242: According to the actualized version of the spec (I know this PR is based on an older proposal), the control flow here is more complex:

- First of all, check whether there is a `permission-required` field at all. If there isn't the rest can indeed be skipped
- If `permission-required` contains `"none"`, skip as well
- Now comes the `hashcash` part
   - To be honest, I'm thinking about prioritizing `hashcash` over `none` in the Rust implementation. Yes, it's more stupid work, but I hope that it makes the transition smoother when the server is under load.
- If none of the above applies, error out (the server only supports permission models we don't know about).

src/wormhole/_rendezvous.py
meejah: Oh, I think that was "note-to-self" about the format .. it should be replaced with a proper docstring for sure :)

src/wormhole/_rendezvous.py
```diff
@@ -61,6 +62,54 @@ def buildProtocol(self, addr):
...
+    import base64
+    while True:
+        attempt = "1:{}:{}:{}::{}:{}".format(
+            bits,
+            timestamp,
+            resource,
+            base64.b64encode(os.urandom(12)),
+            counter,
```
vu3rdd, line 76: It seem like `counter` is also base64 encoded?

meejah: this does seem to work .. it'll be ascii just because it's only ever `[0..9]` ... but maybe instead it should be binary (and then base64 encoded)...?
(The `hashcash` binary that comes with Debian also does a weird thing by default where it lower-cases everything .... not sure why, but it has `-C` to turn that off).

PR #455, opened 2022-10-25
Receive allocate code

### LINKED ISSUE #450: Can't use "wormhole receive" to generate a code
It _looks_ like the intent of the CLI is that ``wormhole receive --code-words 2`` would allocate a 2-length code and await a sender on that mailbox.

However, it doesn't do that: it still prompts for a code.
This means there's no way to use the software in the mode where the receiver generates the code and then communicates that to the sender (where they'd use ``wormhole send --code <whatever>``).

meejah:
Fixes #450 

This allows the receiver to actually allocate a code and give it to the sender.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,8 @@ User-visible changes in "magic-wormhole":
 ## Upcoming Release
 
 * Python 3.5 and 3.6 are past their EOL date and support is dropped (#448)
+* `wormhole receive` now accepts `--allocate-code` so that a sender can
+  use `--code` to send them a file (#450)
 
 ## Release 0.12.0 (04-Apr-2020)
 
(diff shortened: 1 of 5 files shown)


piegamesde:
I think some of these CLI flags should be mutually exclusive if they don't make sense together.

piegamesde:
Yeah my concern is basically that when one enters a code (`wormhole receive 42-foo-bar`), then the other arguments will silently be ignored. But I also see that this is not specific to `--allocate`, but also applies to  `--code-words`. What is the current behavior when receiving with `--code-words`?

meejah:
Yes, `--code-words` only really makes sense with `--allocate` -- but also the default is "2" so most people won't need `--code-words`. So _just_ entering `--code-words` should be an error (although I think the way Click works, `--code-words 2` is precisely the same as not entering it .. maybe there's a way around that).

If you actually enter a code, like `wormhole receive 42-foo-bar` then it should be an error to _also_ request a code with `--allocate`.

meejah:
(p.s. tests are failing due to ... other reasons, there are other PRs for that).

Issue #72, opened 2016-08-04
scale/federate the rendezvous server?


warner:
Looking towards the future: magic-wormhole uses a single central rendezvous server to allow clients to find each other based upon just a small integer. If/when it becomes more popular, the traffic on this server might be an issue. How could we use multiple server hosts to avoid this bottleneck?

(note that this is independent of the "transit relay" server, which is morally equivalent to a TURN server and helps file transfers to work when both clients are behind NAT boxes, by gluing two TCP streams together)

There are two components to the rendezvous server. The first is channel allocation: the most common workflow is for `wormhole send` to allocate a channel by asking the server to assign one, and this needs to be unique (so the server needs to know a complete list of all channel IDs currently in use). These are called "nameplates" in the protocol and the code.

The second is "mailbox" management: a semi-persistent set of messages, and a publish-subscribe protocol for both sides to add and retrieve those messages. Each nameplate maps to a mailbox, each mailbox has a queue/set, and each client connects to a mailbox.

I think the channel-allocation part requires some sort of singular central table. Even if we get multiple servers involved, I think they'll need to talk to a single memcached or something.

The mailboxes could be sharded out among multiple machines, as long as there's a way for every client to get connected to the right machine. For example, we could use DNS round-robin or something to point to a fleet of servers, all of which can use that memcached instance to find out where a given mailbox lives. If a client connects to the wrong server, that server looks up the mailbox in the central table, learns about the correct server, then redirects the client to them. I think that means one non-sharded message per lookup, but everything else scales with the number of servers we use.

stormdragon2976:
I am interested in running a public server. Let me know if I can help out with this. Also, I guess this should probably be a separate issue, but maybe there should be a ~/.config/magic-wormhole/ where per-user configurations could be stored, including preferred relay, etc.

anarcat:
i could have sworn i mentionned this before, but couldn't this be implemented through the [DHT](https://en.wikipedia.org/wiki/Distributed_hash_table)?

there are a few concerns with this:

1. the DHT may not be able to assert unicity of nameplates. could other mechanisms be found for this, for example using UUIDs or something similar?
2. mailboxes may be much more complex to implement and slower

Are you familiar with the DHT?

anarcat:
FWIW, here are a few implementations in Python:

* [Kademlia](https://github.com/bmuller/kademlia): async twisted, follows Kademlia reference implementation, UDP, MIT?
[...]

PR #461, opened 2022-10-30
Parse IPv6 address in TCP hint (--transit-helper argument)

FelisDiligens:
I tried pointing to my own transit-relay server by using `wormhole --transit-helper=tcp:[1111:2222:3333:4444:cccc:dddd:eeee:ffff]:4001 send "./example.file"`, however I got:
```
non-numeric port in TCP hint 'tcp:[1111:2222:3333:4444:cccc:dddd:eeee:ffff]:4001'
```

Not using brackets didn't work either:
```
non-numeric port in TCP hint 'tcp:1111:2222:3333:4444:cccc:dddd:eeee:ffff:4001'
```

I dug around and replaced a line in `parse_hint_argv` in file `_hints.py`.
Once parsing IPv6 addresses worked, I was able to use my transit-relay server to send/receive files over IPv6:

```
Sending (->relay:tcp:1111:2222:3333:4444:cccc:dddd:eeee:ffff:4001)..
100%|██████████████████████████████████████| 13.3k/13.3k [00:00<00:00, 5.00MB/s]
File sent.. waiting for confirmation
Confirmation received. Transfer complete.
```

(An IPv6 address has to be within square brackets, otherwise it isn't recognized.)

It might not be the most elegant solution to use a complicated regex string, but it does the trick. Hostnames, IPv4, and IPv6 addresses are recognized.

### DIFF
--- a/src/wormhole/_hints.py
+++ b/src/wormhole/_hints.py
@@ -50,7 +50,8 @@ def parse_hint_argv(hint, stderr=sys.stderr):
               file=stderr)
         return None
     hint_value = mo.group(2)
-    pieces = hint_value.split(":")
+    # Find all pieces that either match an IPv6 address in square brackets, or pieces that follow or are followed by a colon:
+    pieces = re.findall(r"(?<=\[)[a-fA-F0-9:]+(?=\])|[^:\[\]]+(?=:)|(?<=:)[^:\[\]]+", hint_value)
     if len(pieces) < 2:
         print("unparseable TCP hint (need more colons) '%s'" % (hint, ),
               file=stderr)
--- a/src/wormhole/_hints.py
+++ b/src/wormhole/_hints.py
@@ -40,22 +40,20 @@ def parse_hint_argv(hint, stderr=sys.stderr):
     assert isinstance(hint, type(u""))
     # return tuple or None for an unparseable hint
     priority = 0.0
-    # parse hint type ('tcp' or 'tcp6')
+    # parse hint type
     mo = re.search(r'^([a-zA-Z0-9]+):(.*)$', hint)
     if not mo:
         print("unparseable hint '%s'" % (hint, ), file=stderr)
         return None
     hint_type = mo.group(1)
-    if not hint_type in ("tcp", "tcp6"):
+    if hint_type != "tcp":
         print("unknown hint type '%s' in '%s'" % (hint_type, hint),
               file=stderr)
         return None
     hint_value = mo.group(2)
-    # declare vars
     hint_host = ""
-    hint_port = -1
     pieces = []  # hint_value split into pieces
-    # check for IPv6 address (must have square brackets)
+    # parse IPv6 address (must have square brackets)
     mo = re.search(r'^\[([a-zA-Z0-9:]+)\]:(.*)$', hint_value)
     if mo:
         hint_host = mo.group(1)
@@ -72,17 +70,13 @@ def parse_hint_argv(hint, stderr=sys.stderr):
                   file=stderr)
             return None
         hint_host = pieces[0]
-        if isIPAddress(hint_host) and hint_type == "tcp6":
-            print("expected IPv6 address or hostname, got IPv4 address in TCP6 hint '%s'" % (hint, ),
-                  file=stderr)
-            return None
     # parse the port:
     mo = re.search(r'^(\d+)$', pieces[1])
     if not mo:
         print("non-numeric port in TCP hint '%s'" % (hint, ), file=stderr)
         return None
     hint_port = int(pieces[1])
-    # parse the rest, such as "priority=float"
+    # parse the rest ("priority=float")
     for more in pieces[2:]:
         if more.startswith("priority="):
             more_pieces = more.split("=")
--- a/src/wormhole/_hints.py
+++ b/src/wormhole/_hints.py
@@ -40,28 +40,49 @@ def parse_hint_argv(hint, stderr=sys.stderr):
     assert isinstance(hint, type(u""))
     # return tuple or None for an unparseable hint
     priority = 0.0
+    # parse hint type ('tcp' or 'tcp6')
     mo = re.search(r'^([a-zA-Z0-9]+):(.*)$', hint)
     if not mo:
         print("unparseable hint '%s'" % (hint, ), file=stderr)
         return None
     hint_type = mo.group(1)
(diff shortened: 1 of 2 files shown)


piegamesde:
Thank you for bringing this up. I think the proper thing to do here is to use the proper URL syntax (`tcp://`) here, so that one could offload the parsing to a proper URL parsing library and get all those edge cases for free.

This can be done in a backwards compatible manner like this: [LeastAuthority/wormhole-william@`ab8635f` (#72)](https://github.com/LeastAuthority/wormhole-william/pull/72/commits/ab8635f703b5de3e7adb6cd5420774bccbcb7672)

FelisDiligens:
I made another commit where it passes the parsing to Python's [urlparse](https://docs.python.org/3/library/urllib.parse.html) function, instead of doing the legwork itself.

It also removes 'priority=x.x' before parsing and is backwards-compatible to the old style urls.

I tested a few cases, it seems to work fine:
<details>
<summary>Test cases</summary>

```
# Works:
tcp:transit.magic-wormhole.io:8001
tcp://transit.magic-wormhole.io:8001
tcp:192.168.178.45:4001
tcp://192.168.178.45:4001
tcp:[2001:db8:3333:4444:5555:6666:7777:8888]:4001
tcp://[2001:db8:3333:4444:5555:6666:7777:8888]:4001
tcp://[2001:db8:3333:4444:5555:6666:7777:8888]:4001:priority=2.0
tcp:[::]:4001
tcp://[::]:4001
tcp:[::1]:4001
tcp://[::1]:4001
tcp:example.com:12345:priority=1.0
tcp://example.com:12345:priority=1.0

# Does not work:
tcp:
tcp:example.com
tcp:[2001:db8:3333:4444:5555:6666:7777:8888]
tcp:[2001:db8:3333:4444:5555:6666:7777:8888]:priority=1.0
tcp:2001:db8:3333:4444:5555:6666:7777:8888:4001
tcp::::4001
tcp:::1:4001
tcp://[2001:db8:3333:4444:5555:6666:7777:8888]:4001:priority=xyz
tor:something-something:8001
```

</details>

However, I didn't test it on Python 2.7.

meejah:
It shouldn't need `//` to be a valid URL, lots of things don't (e.g. `mailto:`).

`tcp:` is kind of a weird scheme too (since its transport oriented), if we actually want these to be URLs (instead of transport-specific encodings) they should probably start with `transit:` or something (i.e. related to the actual protocol being spoken, not the transport).

meejah:
In this case, since it _is_ a transport-specific option for a particular program, maybe `tcp6:...` or something makes sense? (And won't e.g. break existing usage like changing to `tcp://` will).

meejah:
Yes, these are just details of this particular CLI implementation. So, the details of what particular URL parsing libraries accept shouldn't affect how the CLI is presented (especially because this _isn't_ a URL).

That said, care does need to be taken not to "bleed out" into the hints (that are sent in the protocol).

"tcp" versus "tcp6" tells the software how to use the information (and Twisted does have different endpoints for v4 versus v6). so `tcp:example.com` can then mean to contact "example.com" with a TCPv4 transport, while `tcp6:example.com` could mean to contact it with v6 transport (while e.g. `tcp6:1.1.1.1` doesn't make sense because it's a v4 address). For hostnames, it affects how you look them up (e.g. A versus AAAA records for example).

meejah:
p.s. I'm actually surprised these _aren't_ Twisted "endpoint strings" since the v4 one looks like it would be a valid one (possibly minus the `priority=` part) ... and maybe they should be (this would mean `tcp6:...` would be the way to support v6 transports, but other things would come "for free" like Tor support by using `tor:...`).

meejah:
> That said, care does need to be taken not to "bleed out" into the hints (that are sent in the protocol).

I just mean here that any peculiarities that this CLI has shouldn't be represented in the hints themselves (e.g. "`tcp6:`" isn't a well-known scheme etc). So, I meant that suggestion for this implementation (only) -- that is, `parse_hint_argv()` is already looking for the string `tcp:` so it could also accept `tcp6:` and treat that as a "must be v6 address" -- although see below about hostnames :)

We shouldn't actually have to care whether the thing is an IP address (of v4 or v6 flavour) or a proper hostname when sending the hint, or parsing it -- the "tcp" hint already has to support all those cases.

So this really is just a bug in https://github.com/magic-wormhole/magic-wormhole/blob/e522a3992217b433ac2c36543bf85c27a7226ed9/src/wormhole/_hints.py#L39 not accepting an IPv6 string properly. So, sorry for not reading the code more closely the first time :(

I think urlparse is probably overkill and maybe regex too? A hostname can't have `[` or `]` in it, right? So that could trigger the "parse this as ipv6, possibly with trailing port" parsing behavior...

FelisDiligens:
Hi, again! Thanks for the quick answer. :)

I pushed two commits that only affect `parse_hint_argv` and `test_parse_hint_argv`.
It can now handle these cases:
```
tcp:[2001:0db8:85a3::8a2e:0370:7334]:1234
tcp6:[2001:0db8:85a3::8a2e:0370:7334]:1234
tcp6:host:1234
```
`tcp6:` is optional and `tcp:` accepts IPv6 addresses too.

It returns `None` in these cases:
```
tcp:[abc::xyz]:1234  # stderr: invalid IPv6 address in TCP hint 'tcp:[abc::xyz]:1234'
tcp6:[abc::xyz]:1234 # stderr: invalid IPv6 address in TCP hint 'tcp6:[abc::xyz]:1234'
tcp6:172.0.0.1:1234  # stderr: expected IPv6 address or hostname, got IPv4 address in TCP6 hint 'tcp6:172.0.0.1:1234'
```

Please let me know, if I missed anything.

meejah:
Great, thanks for the quick follow-up.

It might be good to just get rid of `tcp6` entirely -- that was probably just noise on my part, confusing this with Twisted endpoints etc. :/ 

I don't think the user has any reason to care about that (right?) -- they just want it to work, and TCP is still TCP on v4 or v6...

FelisDiligens:
> I don't think the user has any reason to care about that (right?) -- they just want it to work, and TCP is still TCP on v4 or v6...

I suppose, `tcp:` and `tcp6:` wouldn't make a difference here, so I think you're right. I removed `tcp6:` again and updated the tests.

meejah:
great, thanks!

PR #466, opened 2022-11-11
Use build-in unittest.mock

jelly:
Since Python 3.3 unittest has mock build in and the python-mock module is no longer required.


meejah:
LGTM, do you want to add a short sentence to NEWS.md (please credit yourself too)?

PR #468, opened 2022-12-08
Document magic-wormhole's removal in Fedora 37

Ramblurr:
magic-wormhole was removed in Fedora 37. This PR updates the documentation to reflect that.

* Source 1: https://bugzilla.redhat.com/show_bug.cgi?id=2073777
* Source 2: https://src.fedoraproject.org/rpms/python-magic-wormhole/c/f1352b970b9d289948e137528468ef40bf2e52ba?branch=rawhide

### DIFF
--- a/docs/welcome.md
+++ b/docs/welcome.md
@@ -66,6 +66,10 @@ $ sudo apt install magic-wormhole
 
 ### Linux (Fedora)
 
+Note: magic-wormhole [was removed from
+Fedora](https://bugzilla.redhat.com/show_bug.cgi?id=2073777) starting in Fedora
+37. So this command will only work on Fedora 36 and earlier.
+
 ```
 $ sudo dnf install magic-wormhole
 ```


meejah:
I can't tell _why_ it was removed from those tickets; is it just that there are no Fedora developers interested in packaging it..?

psanford:
This is the fedora bug: https://bugzilla.redhat.com/show_bug.cgi?id=2073777

meejah:
Okay, I see now I guess it's because txtorcon, spake2 and hkdf were removed (the latter is no longer a dependency).

So, I suppose recursive question: why were _those_ removed?

(I tried searching for "txtorcon" and "spake2" which returned no results). I'm not very familiar with Fedora or its development. I am the author of txtorcon too, though, so would be happy to provide anything there -- or here -- that makes sense if there's some reason beyond lack of interest that these are gone.

meejah:
(If someone familiar with Fedora development knows what needs to happen to return magic-wormhole to Fedora please drop a line here or make a pull-request with a note in welcome.md indicating that)

juhp:
I put some comments in the downstream bug, which I hope might help
(sorry probably the comment about the rust port there doesn't help).

meejah:
Yeah, the Rust version doesn't have all the features that the Python client currently does.
Thanks for commenting!

juhp:
(Just to briefly repeat here what I [wrote](https://bugzilla.redhat.com/show_bug.cgi?id=2073777#c7): if someone wants to get magic-wormhole and its 3 deps back into Fedora then they need to be/become a Fedora Packager and take them through the review+unretirement process - not that hard but needs general knownledge of rpm packaging.)

emilazy:
FWIW based on my own experience in Nixpkgs I’m guessing downstreams would be less hesitant about the maintenance status here if there wasn’t a dependency on a cryptography library (`spake2`) that hasn’t been touched in 6 years that depends on a cryptography library (`hkdf`) that hasn’t been touched in a decade. Pure algorithm implementations are maybe the least worrying kind of cryptography library to be completely unmaintained, but it’s still concerning, and nothing else in our large repository (other than optional dependencies on another library for a dead cryptocurrency project) depends on those libraries.

I don’t know if there are other Python implementations of SPAKE2 that could be used instead, though; I found a CFFI‐based binding to a C implementation but I don’t know if that’s really better.

Does the Magic Wormhole project consider the Rust version to be an adequate replacement for the Python one?

meejah:
> Does the Magic Wormhole project consider the Rust version to be an adequate replacement for the Python one? (Asking in case things have changed since 2022.)

No. It lacks important core features still.

meejah:
Note that the library "hkdf" is no longer a dependency, if that helps.

"spake2" is still (but is maintained by the original author of magic-wormhole)

emilazy:
`spake2` itself depends on `hkdf`, doesn’t it? I see you’ve opened a pull request to move over to `cryptography` at https://github.com/warner/python-spake2/pull/14, which would be great. However, considering that that pull request has been open for 2 years now and `spake2` also has [issues on Python 3.12](https://github.com/warner/python-spake2/pull/16), I am somewhat concerned about the maintenance state of the library for being a core cryptographic component of Magic Wormhole. (I don’t mean this as a criticism of the maintainer, of course; I love Magic Wormhole, and everyone has limited time.)

meejah:
Yes, my bad (edited above comment). I believe I was confused about PR 14 indeed :)

emilazy:
I do realize there aren’t necessarily good options here and just because libraries haven’t been touched in many years doesn’t mean they’re vulnerable, so I appreciate your position here, and we might just go down the path of fixing `hkdf` (and `spake2` if necessary). But I wonder if it might be worth at least forking `spake2` into the `magic-wormhole` organization so that changes like the `cryptography` migration and Python 3.12 fixes could be integrated?

meejah:
I have contacted Brian about spake2, so should be able to make progress on it soon (and yes, perhaps moving it to this org would be better overall)

meejah:
p.s. I do think dropping `hkdf` is good overall (and in the very short-term): there _is_ a good alternative there (in "cryptography") so it should be preferred.

emilazy:
Thanks for listening to our pesky downstream concerns :) Getting `spake2` actively maintained again and ditching `hkdf` sounds like a great outcome to me! I don’t know if it’ll make a difference for Fedora specifically, but it sounds like it’ll be good for the ecosystem in general and Nixpkgs specifically. It’d be helpful for us if you could post any updates on this issue, but otherwise I’ll check in on the state of things in a bit.

meejah:
@emilazy in case you didn't already get a bunch of notifications, I've opened a relevant issue to improve the dependencies issue (#534). Thanks!

meejah:
for the record, HKDF is no longer a python-spake2 dependency (as of ~7 months ago)

PR #477, opened 2023-04-05
476.special characters

### LINKED ISSUE #476: Security Vulnerability: Control characters in filename can manipulate terminal of receiver
When magic-wormhole displays filenames, control characters in the file are written to stdout verbatim. This allows attackers to manipulate the terminal environment of the receiving party. There is a multitude of possible attacks, but the simplest is to overwrite existing output of magic-wormhole and thus manipulate file size and name of the transferred party.

## Proof of Concept
Sender:

    touch "$(echo -e 'MaliciousPrefix\rReceiving file (351 Bytes) into: completely_harmless.txt')"
    wormhole send MaliciousPrefix^MReceiving\ file\ \(351\ Bytes\)\ into\:\ completely_harmless.txt

Receiver:

    $ wormhole receive […]
    Receiving file (351 Bytes) into: completely_harmless.txt
    ok? (Y/n): 
    Receiving (->tcp:192.168.0.2:34411)..
    0.00B [00:00, ?B/s]
    Receiving file (351 Bytes) into: completely_harmless.txt
    
The receiver does not see the real file name or the real file size, but instead gets presented with attacker controlled text.

## Workaround
Wrap the output of wormhole receive so that control characters are masked. E.g.:

    $ wormhole receive 21-midsummer-endow 2>&1 | cat --show-nonprinting
    Receiving file (0 Bytes) into: MaliciousPrefix^MReceiving file (351 Bytes) into: completely_harmless.txt
    ok? (Y/n): 
    Receiving (->tcp:192.168.0.2:45475)..
    ^M0.00B [00:00, ?B/s]^M0.00B [00:00, ?B/s]
    Received file written to MaliciousPrefix^MReceiving file (351 Bytes) into: completely_harmless.txt

## Fix
Escape all control characters (and other untypical characters to be sure) so that injection attacks are no longer possible. Simplest solution is to `repr()` all attacker controlled strings, which will also put quotation marks around it. Alternatively you could use a terminal library which correctly embeds arbitrary byte strings into the context.

meejah:
Fixes #476

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,7 @@ User-visible changes in "magic-wormhole":
 * Use the HKDF primitive from "cryptography" (#462)
 * Replace mock module with Python's builtin unittest.mock
 * Stream to disk after 10MB on directory receive (#447)
+* SECURITY: Replace "weird" characters in receiver's display (#476)
 
 Thanks to the many contributors of bugs, patches, and other help with this
 release:
--- a/src/wormhole/cli/cmd_receive.py
+++ b/src/wormhole/cli/cmd_receive.py
@@ -18,7 +18,7 @@ from wormhole import __version__, create, input_with_completion
 from ..errors import TransferError
 from ..transit import TransitReceiver
 from ..util import (bytes_to_dict, bytes_to_hexstr, dict_to_bytes,
-                    estimate_free_space)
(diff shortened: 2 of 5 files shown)


meejah:
Going with the "simple" approach of "just using `repr()`"

PR #490, opened 2023-05-31
remove add_relay event handling from the state machine

### LINKED ISSUE #489: add_relay state machine input event is not used
All the connection hints (direct and relay) is sent with the `connection-hints` type message. The `manager` module responds to these `rx_HINTS` events and sends them to the connector via  `got_hints` event. There is some [stale code](https://github.com/magic-wormhole/magic-wormhole/blob/master/src/wormhole/_dilation/connector.py#L154) that deals with `add_relay` event that can be removed.

vu3rdd:
The associated test code has been removed as well.

Closes #489

### DIFF
--- a/src/wormhole/_dilation/connector.py
+++ b/src/wormhole/_dilation/connector.py
@@ -148,12 +148,6 @@ class Connector(object):
 
     # TODO: unify the tense of these method-name verbs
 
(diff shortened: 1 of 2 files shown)


meejah (approved):
LGTM!

(Waiting for CI before merging).

Issue #58, opened 2016-06-23
Do not create zipfile when sending much data


bofm:
Sometimes people have less free space on disk than the size of the zipped data they want to send. In such cases it would be reasonable to stream data and calculate checksums on the fly rather than to create a zipfile first.

warner:
Hm, good idea. There's a subtle failure mode we have to be careful about, though, where the transfer gets interrupted partway through. Basically we just need to stream into a temporary directory and their either rename it into place when it succeeds, or delete the fragment if it fails.

I'll look into the zipfile module and see if there's an easy way to stream data into it. It might want to do `seek()` on the input stream, which would make it pretty tricky.

warner:
Hm, maybe we need to launch `tar` or `zip` in a subprocess, and stream its data to the other side. We wouldn't know how large the zipfile will be ahead of time, though, so the best we could do to advise the receiver about what they're about to get would be to run `du` over the source directory. We'd also need to make sure to not allow a truncated transfer to look like a complete one (the current code knows exactly how many bytes to expect; the new implementation would need to send a special distinct "stop" message at the end).

I suspect subprocesses won't work too well on windows (well, and it's not like we can expect `tar` to be there), so this mode might fail if either side is on an OS that lacks these tools.

bofm:
No need to do subprocess. Python has both tar and zip support in the standard library. And the size of the whole directory can be calculated in Python combining `os.walk` and `os.stat`.

warner:
Yeah, but we can only use the stdlib tar/zip libraries if there's a way to make them streaming-friendly:

* On `send`, we need to create a fake file-like object to pass into `zipfile.ZipFile()` for writing. Each time the zipfile object calls `obj.write(data)`, we need to send that data over the wire. If the zipfile object ever calls `obj.seek()`, then we're hosed (and I assume it will want to seek back to the beginning, after writing out all the data, to fill in the directory table).
* On `receive`, the `ZipFile` will be calling `obj.read(length)` on our file-like object, which needs to return the compressed data. But the flow control is going in the other direction: the receiver reacts to new data arriving, rather than asking for data to be input. So we'd either have to run the `ZipFile` in a thread somehow, or write the whole stream to a tempfile (so we can be sure it's all available before `obj.read()` gets called), which uses up just as much disk space as before.

Running them in a subprocess would let us use stdin/stdout as the file-like object, and deal with the flow-control in the pipe that sits between them.

[...]

PR #503, opened 2023-10-02
Implement streaming zipfile creation for sending directories

pR0Ps:
Instead of creating a temporary zip file on disk to send when transferring a directory, this commit uses the [zipstream-ng](https://github.com/pR0Ps/zipstream-ng) library to generate the zip file on the fly. This results in no disk usage, a much shorter startup time, and minimal memory usage.

For adding files to the stream from the directory, the `walk` function the `zipstream-ng` package provides was used since it preserves empty directories (obsoletes #422) as well as follows symlinks. Since files aren't actually read until they're about to be transferred, an `os.access` check is done to make sure that unreadable files are not added to the stream (which would cause it to crash mid-transfer).

I tested sending ~90GB and the transfer started immediately and finished successfully with minimal memory usage. Transfer speed was about the same as sending a non-zipped file.

(This is a cleaned up implementation of the proof of concept originally proposed here: https://github.com/magic-wormhole/magic-wormhole/issues/58#issuecomment-1672468894)

### DIFF
--- a/src/wormhole/cli/cmd_send.py
+++ b/src/wormhole/cli/cmd_send.py
@@ -1,12 +1,11 @@
 from __future__ import print_function
 
+import errno
 import hashlib
 import os
 import sys
 
 import stat
-import tempfile
-import zipfile
 
 import six
 from humanize import naturalsize
@@ -21,6 +20,9 @@ from ..errors import TransferError, UnsendableFileError
 from ..transit import TransitSender
 from ..util import bytes_to_dict, bytes_to_hexstr, dict_to_bytes
 from .welcome import handle_welcome
+
+from iterableio import open_iterable
+from zipstream.ng import ZipStream, walk
 
 APPID = u"lothar.com/wormhole/text-or-file-xfer"
 VERIFY_TIMER = float(os.environ.get("_MAGIC_WORMHOLE_TEST_VERIFY_TIMER", 1.0))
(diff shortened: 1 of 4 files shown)


meejah:
This is looking really nice, thanks for the PR! :partying_face: 

My only minor hesitation is that it depends on two relatively-new libraries and I don't know what implications that may have for downstream packaging. (I haven't looked too closely at the libraries yet either).

Do you know if zipstream is in e.g. Debian or any distros yet?

Thanks again for your work on this, streaming large directories w/o using a ton of disk is a nice feature for sure!

pR0Ps:
I'm the maintainer of both of the added libraries (sorry, I had a disclaimer in the original comment but didn't copy it into the PR).

[`zipstream-ng`](https://github.com/pR0Ps/zipstream-ng) is currently packaged in Debian (https://packages.debian.org/source/trixie/python-zipstream-ng) (though not the latest version yet), but [`iterable-io`](https://github.com/pR0Ps/iterable-io) is not.

`iterable-io` is extremely small though (about 70LoC without comments) and can just be vendored into the code if packaging would be an issue. I did actually have it just copied in, but ended up extracting it out into it's own library since I had a couple of other unrelated projects that needed to use its functionality too.

meejah:
Cool, thanks.
Great that the "bigger" dependency is already packaged :)

(Also, although I'm happy to do my best to make packaging easier, "making new software available" is what distros have chosen to do, so it's their problem overall).

Vendoring that lib would be worst-case, IMO -- better to let you continue to maintain it :)

src/wormhole/cli/cmd_send.py
```diff
@@ -335,61 +337,41 @@ def _build_offer(self):
...
+                        raise UnsendableFileError(errmsg)
+                    print(
+                        "{} (ignoring error)".format(errmsg),
+                        file=args.stderr
+                    )
+
+            filesizes = [x["size"] for x in zs.info_list() if not x["is_dir"]]
+            filesize = len(zs)
             offer["directory"] = {
                 "mode": "zipfile/deflated",
```
pR0Ps, line 364: Note: I left this even though it's not technically correct anymore. `"zipfile/stored"` would be more accurate since compression is no longer used. I left it because changing it would break compatibility with existing clients that don't know how to handle `"zipfile/stored"` and it doesn't really matter if compression was used while extracting it. Open to alternative suggestions.

meejah: That sounds like the right choice to me.

piegamesde: So I checked what the other clients currently do. Wormhole William does not check the value at all, it just assumes it is a zip file and unzips it. wormhole-rs doesn't check either, but it doesn't need to because it passes on the compressed file without decompressing it for the user, so any format would be "fine". The only implementation which would not accept it is the Python code, which actually does check the value.

I do think that `zipfile/deflated` is very unfortunate to begin with for the reasons you mention, but also that `zipfile/stored` wouldn't be any better for the same reason. Ideally, the value should be simply `zipfile`. Of course this is too late to change now and therefore using `zipfile/deflated` is the best thing to do currently. But I think at the very least that the check in the Python version should be softened to also accepting `zipfile`, so that we may hopefully be able to migrate some time down in the future.

https://github.com/magic-wormhole/magic-wormhole/blob/master/src/wormhole/cli/cmd_receive.py#L332

pR0Ps: @piegamesde implemented in the latest commit

meejah:
Sorry for the delay .. trying to understand the implications of the last change and "deflate" expectations. Tested against Warp (but it doesn't auto-unpack anyway it seems).

pR0Ps:
The last commit only makes the `"mode"` check on incoming directories a bit looser so that it will accept just `"zipfile"` instead of only `"zipfile/deflated"`. For sending data, the "mode" is unchanged since it's expected that other clients (or just older magic-wormhole versions) would be relying on the sent mode being the same `"zipfile/deflate"` as it always has been.

The other change was removing the `compression` parameter from the `zipfile.ZipFile(...)` call that reads the incoming zip. This is essentially a no-op because the zip is being opened in read-only mode and the specified compression value is only ever used for writing files (you can test this by creating a zipfile using `compression=ZIP_LZMA` and reading it back with `ZipFile(..., mode="r", compression=ZIP_DEFLATED)` - it still works).

meejah:
Thanks again for the PR, sorry for the delays reviewing + merging!

pR0Ps:
Thanks for merging this! I think #422 and #232 can be closed now since they're about preserving empty directories which this PR added.

acrefoot:
This works beautifully on an OS X (M1 mac studio)! Can't wait for it to be cut as a release in PyPI.

meejah:
Great, always good to get test-reports (successful or not!)

PR #505, opened 2023-10-24
Update versioneer to 0.29

aduskett:
Fixes builds against Python 3.12.0

### DIFF
--- a/versioneer.py
+++ b/versioneer.py
@@ -1,5 +1,4 @@
-
-# Version: 0.18
+# Version: 0.29
 
 """The Versioneer - like a rocketeer, but for versions.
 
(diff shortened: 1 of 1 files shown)


meejah:
Is there an easy way for me to confirm this is versioneer 0.29?

(I'm absolutely not assuming anything malicious, but I've got at least 4 of these sorts of PRs to look at right now and would like to do better than "stare meaningfully at the diff" ;) )

meejah:
Can you say how you created this?

If I run `versioneer install --vendor` I get different results (mostly looks like "formatting stuff").

meejah (changes requested):
I get different results from `versioneer install --vendor` with 0.29 -- the next upgrade will be easier if there's no formatting differences, IMO.

If you do that please also add an entry in NEWS (include your username or similar).

aduskett:
No worries! Sorry, I was updating the packages for The Buildroot embedded SDK project and there were several packages using ancient versions of versioneer and by the 20th hour I didn't have the energy to make all the pull requests valid. :)

meejah:
No worries .. it must have been some local formatting thing? (The changes were all boring but I committed the output from versioneer directly so the 0.30 upgrade is less surprising ;)

PR #525, opened 2024-04-25
Update wormhole_complete.bash - fix #524 

### LINKED ISSUE #524: Error "-bash: complete: nosort: invalid option name" on macOS, possibly due to bash version
Installed magic-wormhole on macos with homebrew:
`brew install magic-wormhole`

If bash is the default shell, the error "-bash: complete: nosort: invalid option name" appears at terminal startup.

This issue seems to be caused by the `-o nosort` option in https://github.com/magic-wormhole/magic-wormhole/blob/241a8358ce8ec4e12fd393f562e9dbee6484eba9/wormhole_complete.bash#L25 located in `/usr/local/etc/bash_completion.d/wormhole.bash`

RobertoD91:
fix for #524

### DIFF
--- a/wormhole_complete.bash
+++ b/wormhole_complete.bash
@@ -22,7 +22,11 @@ _wormhole_completion() {
 }
 
 _wormhole_completion_setup() {
-    complete -o nosort -F _wormhole_completion wormhole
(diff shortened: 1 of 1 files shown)


meejah:
Oh, hrmm ... every time we "re-build bash completions" I think it'll get reverted....

PR #526, opened 2024-05-24
Add codespell support (config, workflow to detect/not fix) and make it fix few typos

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -1,0 +1,25 @@
+# Codespell configuration is within setup.cfg
+---
+name: Codespell
+
+on:
+  push:
+    branches: [master]
+  pull_request:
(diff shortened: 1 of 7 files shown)


yarikoptic:
sure, we can whitelist any word ... whitelisted that one and specific cased `assertIn` (so that `assertin` would still be thought to possibly be a typo of  `assertion` I guess)

meejah:
Okay, great.

This looks like it found a few misspellings and so I'm happy to try it -- we can always remove the CI etc later if it for some reason proves to be more-annoying than worth it :) 

thanks!

PR #546, opened 2024-10-15
display QR code when sending

Ferada:
Implements the QR code display from <https://github.com/magic-wormhole/magic-wormhole/issues/431> - though I went with `--qrcode`, can change it though if you think it's too long.

Tried with some Android application, works fine.

I was wondering about the zero mode, probably doesn't make sense to have the QR code for that, right?

Lastly, screenshot:

```
(.venv) 0 magic-wormhole % wormhole send NEWS.md
Sending 29.3 kB file named 'NEWS.md'
Wormhole code is: 4-infancy-classroom
On the other computer, please run:

wormhole receive 4-infancy-classroom



    █▀▀▀▀▀█   ▄▀▀▀▄▄█▄▀▀▀ █▀▀▀▀▀█
    █ ███ █ ████▄ ▀ ▀▄▀█▀ █ ███ █
    █ ▀▀▀ █ █▀█▀██▄▄█ █▀█ █ ▀▀▀ █
    ▀▀▀▀▀▀▀ █ █ █▄▀ █ ▀ ▀ ▀▀▀▀▀▀▀
    █▄█▀█▀▀▄▄▀▀▄▀ ▀ ▄█▄█▄▄█████▄▄
    ▄ ▀▄▀▀▀█▄▄█▄  ▄▀ ▀▄ █▄█▀██ ▄▄
    ▄ ▄  ▀▀███▄▄ ▄▄▀█▀▄█  ▄▄▄ █▀▄
    ▀███▄█▀ ▀█▄█▀█▀▀▀▀▀ ▄▀▄ ▄▄  ▄
    ▄▄ ▄▀▀▀ ▀▀█▄▀ ▀▀██ ▄▀▄▄█▄▄█▄█
    █ ▄   ▀▄▀▄▀▄▀ ▄▀▀ ▀▄ ▄▄▄██▀ ▄
    ▀ ▀▀▀▀▀▀█▀   ▄▄▀█████▀▀▀█▄█ ▄
    █▀▀▀▀▀█ ▄▀ ▀██▀█▀  ▀█ ▀ █▄
    █ ███ █ ██ ▀  ▀▀▄▀▄ █▀▀█▀▄▀▄▄
    █ ▀▀▀ █ ▀█▀▄█ ▄█▄▀▀▄▄▀▀██▀ █
    ▀▀▀▀▀▀▀ ▀▀ ▀ ▀   ▀ ▀   ▀  ▀



```

PR #555, opened 2024-11-13
ferada: display QR code

meejah:
Same as https://github.com/magic-wormhole/magic-wormhole/pull/546 with some test fixes and minor tweaks.


meejah:
Here is what this would look like if the default is "show a QR code"

wormhole-with-qr

meejah:
I'm heavily leaning torwards "QR code by default". Those who care can put `--no-qr` to have the old default.

Ferada:
Awesome, thanks for your help with this!

Issue #558, opened 2024-11-16
api-example.py requires small changes to run out of the box


OndrewBot:
Hi there, I'm familiarizing myself with Magic Wormhole and was starting with api-example.py. It doesn't run right away due to the reassignment of  `relay_url = "ws://localhost:4000/v1"`, which is easy enough to comment out and let the `public_relay.RENDEZVOUS_RELAY` do it's thing. But I'm wondering if some friendly improvements can be made for this?

Another example of a potential mod is that I expected it to run similar to the CLI `wormhole send` and `wormhole receive`, but the `example_initiator` is the one actually receiving the message - totally valid but surprising.  The `async def go()` is obviously intended for that but missing some pieces. 

## Edit Summary
- comment out a relay_url and explain
- finish building out go_outbound()/go_inbound() or change the existing examples to something that matches the CLI send/receive pattern
##
I'm looking to contribute if these changes are worth it to you!
_[proposed labels]: easy, enhancement_

meejah:
Hello!

I'm not sure _exactly_ what that was intended to illustrate, but yes your idea to make it look a lot more like a "send" / "receive" pair would be great!

Let me know if you need any help :) 

(Other examples could also be useful, of course, but I like the above idea for sure)

OndrewBot:
Ok great, I'll open a PR. 
I think I'll just keep updates straightforward for now and refactor the examples to follow that traditional send/receive pattern. I don't want to lose everything that's there now though so I'll see if I can't find a balance comments/instructions/other examples.

meejah:
Oh, I forgot to say in the PR: you can add to `NEWS.txt` when doing PRs if you want :)

Thanks again for following up with a PR

PR #562, opened 2024-12-06
Refactor api-example.py for Issue #558

OndrewBot:
Changes made to let api-example.py run out of the box and have it act like the `wormhole send` `wormhole receive` pattern.

- `example_initiator()` is now `example_sender()`. It will send out a message to the receiver.
- `example_responder()` is now `example_receiver()`. It will receive and print out the whole message now, not just byte length
- moved the extra websocket `relay_url` from out of the functions and to the top as an example 
- moved comments around (kept the wording as much as possible) to fit the new pattern
- added some intro instructions at the top:

```python
# Run api-example.py in 2 different processes:
#   1 -- entering 0 arguments will initiate a wormhole send with a hidden msg
#   2 -- entering the sender's code will inititiate a wormhole receive

# The api can be set up to use a different relay like a websocket
# relay_url = "ws://localhost:4000/v1"
``` 

Please see the rendered documentation: https://magic-wormhole--562.org.readthedocs.build/en/562/

### DIFF
--- a/docs/api-example.py
+++ b/docs/api-example.py
@@ -4,6 +4,7 @@ from wormhole.cli import public_relay
 from twisted.internet.defer import ensureDeferred
 from twisted.internet.task import react
 
+# To run this api locally try: relay_url = "ws://localhost:4000/v1"
 
 async def go():
     w = wormhole.create(appid, relay_url, reactor)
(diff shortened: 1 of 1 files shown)


docs/api-example.py
```diff
@@ -4,67 +4,68 @@
...
+    # once we get the "versions" message we have a viable
+    # communication channel with shared encryption key
+    versions = await w.get_versions()
 
-    msg = await w.get_message() # gets exactly one message
-    print(f"got msg: {len(msg)} bytes")
+    # our simple protocol is: the responder sends a single
+    # message. the flow of messages after "versions" is exchanged is
+    # up to application requirements
+    w.send_message(b"privacy is a human right")
```
meejah, line 36: :)

 :+1:

meejah (approved):
LGTM thanks!

PR #564, opened 2024-12-09
Stop using twisted.internet.defer.returnValue

cjwatson:
`defer.returnValue` was only needed in Python 2; in Python 3, a simple `return` is fine.

`twisted.internet.defer.returnValue` is deprecated as of Twisted 24.7.0.


meejah:
Hrm, looks like a not-related problem with CI for Docs builder.

meejah (approved):
:+1:

meejah:
Great, thanks!

PR #570, opened 2025-01-25
Add WORMHOLE_ACCEPT_FILE environment variable

### LINKED ISSUE #569: Add environment variable for `wormhole receive`'s `--accept-file` flag
Similar to `WORMHOLE_RELAY_URL` and `WORMHOLE_TRANSIT_HELPER`, it would be nice if the `--accept-file` flag on receive operations could be set with an env var like `WORMHOLE_ACCEPT_FILE`.

Being able to set this option in the shell would improve the UX for (especially recurring transfers) where the user(s) know they want to accept every file.

Current flow:

1. Run `wormhole send $file` on the sender side.
2. Copy the `wormhole receive` command with code.
3. Paste and run `wormhole receive` command with code.
4. Wait for "ok? (Y/n):" prompt and input "y" + Enter key.
5. File is transferred.

With `WORMHOLE_ACCEPT_FILE` set:

1. Run `wormhole send $file` on the sender side.
2. Copy the `wormhole receive` command with code.
3. Paste and run `wormhole receive` command with code.
4. File is transferred.

At least for the single-user use case I think this would make a lot of sense to have, and in the multi-user case too I think it would be better (from a security perspective) to have the receiving side explicitly configure this, rather than modify the `wormhole send` command output on the sender side to include the `--accept-file` flag in its output.

haatveit:
If set to a string value that Click considers truthy, file transfers will be accepted without confirmation. https://click.palletsprojects.com/en/stable/parameters/#parameter-types

This commit resolves #569.

### DIFF
--- a/src/wormhole/cli/cli.py
+++ b/src/wormhole/cli/cli.py
@@ -308,6 +308,7 @@ def go(f, cfg):
 @click.option(
     "--accept-file",
     is_flag=True,
+    envvar="WORMHOLE_ACCEPT_FILE",
     help="accept file transfer without asking for confirmation",
 )
 @click.option(
--- a/src/wormhole/test/test_args.py
+++ b/src/wormhole/test/test_args.py
@@ -191,6 +191,11 @@ class Receive(unittest.TestCase):
             cfg = config("--transit-helper", transit_url_2, "receive")
         self.assertEqual(cfg.transit_helper, transit_url_2)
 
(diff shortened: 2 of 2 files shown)


meejah (approved):
LGTM

meejah:
I was going to suggest adding some documentation for this, but the CLI documentation isn't quite up to that task so I filed https://github.com/magic-wormhole/magic-wormhole/issues/571

PR #572, opened 2025-01-29
Documentation edits

meejah:
Please see the rendered documentation: https://magic-wormhole--572.org.readthedocs.build/en/572/


shapr (approved):
This is better, I approve.

meejah:
Okay, now I remember: without the "Pillow<10" restriction, the builds "works" as in nothing is red, but the RtD build doesn't produce any of the seqdiag output (e.g. https://magic-wormhole--572.org.readthedocs.build/en/572/server-protocol.html#diagram-of-normal-interaction ) whether it is inline or from files.

meejah:
These are rendered as "warnings" even though it means there no actual output :( 

```
WARNING: dot code 'seqdiag {\n    peer -> server [label = "type=bind\\nappid=demo\\nside=b491c"];\n}': 'ImageDraw' object has no attribute 'textsize'
```

So  https://github.com/python-pillow/Pillow/issues/7523 says "yeah, it's gone" so I guess really a seqdiag / blockdiag bug? There is a fork which fixes this: https://github.com/yuzutech/blockdiag/

Issue #575, opened 2025-02-20
Better feedback when receiving side fails


meejah:
@shapr and I had an interesting failure mode:

- shae sends a file
- meejah tried to use the "Wormhole" mobile application (on GrapheneOS)
- it consumed the code, but error'd on failing to open a local file
- (shae's side didn't show anything)
- meejah then tried to receive on the command-line (with the same code)
- nothing happens

If the code is "already consumed", meejah's second try should have shown "crowded". Instead it is doing nothing while saying "waiting for sender".

So somehow the receiver + sender are out of sync, but not showing anything useful to their humans.
See also https://gitlab.com/lukas-heiligenbrunner/wormhole/-/issues/35

meejah:
Trying again from scratch worked. However, a "feedback" improvement: when Meejah had connected, but not yet answered "Y/n" to actually receive the file, Shae's side hadn't yet shown any additional feedback.

I think it should be able to say something useful (like "the peer has connected to the mailbox" or similar).

meejah:
Can repeat the original bug with just the Python CLI:

- meejah does `wormhole receive` with the correct code
- when asked "Y/n", meejah did `ctrl-z` and kill the program (e.g. maybe it crashed here? kind-of like what the phone app did?)
- shae received no feedback
- meejah tries again, with the same code
- shae receives no feedback
- meejah's side says "waiting for sender" and does nothing else
- the transfer doesn't work

I would have expected to see an "error: crowded" when I tried a second time (since I've consumed the code).

meejah:
Ahhh, @shapr and I figured it out while staring at state-machine diagrams ;)

What happens when meejah "re-tries" is that the Nameplate for the transfer 1 has already been de-allocated, and so it's not a "re-try" but instead a completely new transfer attempt with "receiver going first" -- so the same nameplate from transfer 1 gets re-allocated for transfer 2, but pointing at a now-different Mailbox. So it's not "crowded" because it's a new Mailbox.

So this also is a "user feedback" issue. It would probably already be more obvious if we told the user (on the sender side) when "mailbox communication has been achieved".

PR #576, opened 2025-02-20
morally correct but imperfect nameplate deallocation notification

shapr:
This is a morally correct but imperfect PR for #575 
From talking to @meejah I think it's likely I need to get stdout from the `Config` object somehow, so tests can check the output.

### DIFF
--- a/src/wormhole/_nameplate.py
+++ b/src/wormhole/_nameplate.py
@@ -156,6 +156,8 @@ class Nameplate(object):
 
     @m.output()
     def T_nameplate_done(self):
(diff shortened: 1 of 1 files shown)


meejah:
...so if we can show on the sending side "the mailbox is now consumed" when the nameplate is de-allocated, we may be able to give better feedback on both sides. Almost certainly wants #591 to land first.

meejah:
Extended this to use the status API -- needed some (internal) refactoring of the status API, and a new public item to do with the "code status".

https://github.com/magic-wormhole/magic-wormhole/pull/599

PR #579, opened 2025-02-21
Add minimal pyproject.toml

### LINKED ISSUE #577: build system upgrades required
When installing a local checkout with new-ish tools, one will see a message like this:

>   DEPRECATION: Legacy editable install of magic-wormhole==0.17.0+79.gbd10148 from file:///home/meejah/src/magic-wormhole (setup.py develop) is deprecated. pip 25.0 will enforce this behaviour change. A possible replacement is to add a pyproject.toml or enable --use-pep517, and use setuptools >= 64. If the resulting installation is not behaving as expected, try using --config-settings editable_mode=compat. Please consult the setuptools documentation for more information. Discussion can be found at https://github.com/pypa/pip/issues/11457

Since installing via something like `pip install --editable .` is a common developer workflow, we should fix this.

I think the "bare minimum" (via @SnoopJ on `#python`) is declaring that setuptools is our build system in a `pyproject.toml`. We could also get rid of `setup.py` entirely (e.g. by switching to "hatch" or one of the other newer build systems).

SnoopJ:
Closes #577.

This changeset adds a minimal `pyproject.toml` the _only_ declares the build requirements and that `setuptools` is the build backend.

Note that I was not able to reproduce the deprecation warning described in #577 (I tried a few combinations of `pip` version, `setuptools` version, etc.), but based on the linked upstream issue I'm pretty confident this is the resolution. And even if it isn't, it's standards-compliant.

The lower bound on the `setuptools` version is meant to exclude versions of `setuptools` that don't support PEP 660.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,3 @@
+[build-system]
+requires = ["setuptools>=64.0.0", "wheel"]
+build-backend = "setuptools.build_meta"


meejah (approved):
LGTM!

Thanks! (waiting for CI to run)

SnoopJ:
Oops, the CI failures are because `versioneer` also counts as a build-time req.

meejah:
:tada: thanks

PR #584, opened 2025-03-03
Fix tests with Click 8.2

cjwatson:
https://github.com/pallets/click/pull/2523 introduced changes to `click.testing.Result` that broke a couple of unit tests in magic-wormhole.  Although this Click version hasn't been fully released yet, this adjusts magic-wormhole to work with both old and new versions.


meejah:
Great, thanks!

PR #590, opened 2025-03-07
Reconnect timeout

meejah:
Makes the implementation match the documentation: adds a (default 30s) timer that sends Pings and demands at least one Pong inside of two timeout interval -- that is, if we haven't seen a Pong for 60s the Leader will consider the connection dead and trigger a new generation.

**DRAFT:** because this is stacked on top of the "dilation feedback experiments" branch still.

Please see the rendered documentation: https://magic-wormhole--590.org.readthedocs.build/en/590/

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,9 @@
 default:
 	echo "see Makefile"
 
+lint:
+	tox -e flake8less
+
 completions:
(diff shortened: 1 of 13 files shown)


meejah:
reviewed via https://github.com/meejah/magic-wormhole/pull/1 (so it could diff again the branch it was stacked on)

PR #591, opened 2025-03-07
Dilation + Wormhole status / feedback

meejah:
Add a "status" API that delivers relevant Wormhole (and possibly also Dilation, if you call `dilate()`) information. Intended for use by client code (e.g. GUI, TUI) to show their human better information about what the wormhole is doing.


meejah:
see also https://github.com/magic-wormhole/magic-wormhole/issues/592

shapr (approved):
crikey what a beaut!

PR #599, opened 2025-03-17
Notify user when the code is consumed

meejah:
See description: https://github.com/magic-wormhole/magic-wormhole/pull/576#issuecomment-2721681884

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,9 @@ default:
 lint:
 	tox -e flake8less
 
+test:
+	python -m pytest -W ignore::DeprecationWarning -r w -v -s src/wormhole/test
+
 completions:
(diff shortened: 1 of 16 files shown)


meejah:
I do not like the "status" tests anymore, they should be refactored too.

meejah:
(This took a bit more refactoring than I hoped, but I think the end-result is better for the status API)

shapr (approved):
I approve of this message, especially the nifty union types used for checking event ordering.

meejah:
I added some clarifying comments about the ordering. My conclusion is that `Union[A, B, C]` or `A | B | C` will retain order -- _unless_ you also put it in a generic

PR #604, opened 2025-03-26
fix error message in title of readthedocs

shapr:
Previously the title tag of the readthedocs site included the string:

`Failed to load config from /home/docs/checkouts/readthedocs.org/user_builds/magic-wormhole/checkouts/latest/pyproject.toml` 

I tracked that down to a line in [versioneer.py](https://github.com/magic-wormhole/magic-wormhole/blob/master/versioneer.py#L410) and added the [missing line](https://github.com/python-versioneer/python-versioneer/blob/master/INSTALL.md#common-steps) to `pyproject.toml`

I tried to choose sensible values for the config options. 
This does change the version displayed in the docs to whatever is extracted from git. 
In local tests, I got this value: `Magic-Wormhole 0.18.0+154.gadc01dd.dirty`.

I would appreciate any feedback on this PR! Is versioneer even used by magic-wormhole?

Please see the rendered documentation: https://magic-wormhole--604.org.readthedocs.build/en/604/

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -58,9 +58,9 @@ Delegated mode:
 
    class MyDelegate:
        def wormhole_got_code(self, code):
-	   print("code: %s" % code)
+           print("code: %s" % code)
        def wormhole_got_message(self, msg): # called for each message
-	   print("got data, %d bytes" % len(msg))
+           print("got data, %d bytes" % len(msg))
 
    w = wormhole.create(appid, relay_url, reactor, delegate=MyDelegate())
    w.allocate_code()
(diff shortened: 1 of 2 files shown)


docs/api.rst
```diff
@@ -58,9 +58,9 @@ Delegated mode:
 
    class MyDelegate:
        def wormhole_got_code(self, code):
-           print("code: %s" % code)
+	   print("code: %s" % code)
```
hacklschorsch, line 61: This reminds me of something

tabs-spaces-both

hacklschorsch: Good to see you around @shapr ❤️

meejah (commented):
I think we only need the pyproject.toml changes (right?)

meejah (approved):

PR #614, opened 2025-04-24
search for text mentions of .md files, replace with refs

shapr:
I used the pattern of what appears to be a ref link in one RST file, hopefully this is the right thing.

Please see the rendered documentation: https://magic-wormhole--614.org.readthedocs.build/en/614/


shapr:
Yup, it does the right thing!

PR #621, opened 2025-05-12
Declarative build

meejah:
"Minimal" changes to get a declarative build (i.e. still use versioneer, setuptools)


gdt:
I think this helps with the extra files at top-level.  I couldn't just use the patch, as master has changed a lot since the release.  But I just built this in pkgsrc and it seems ok.

meejah:
great, thanks for the feedback!

PR #617, opened 2025-05-02
Explicit subprotocol interface for subchannels

meejah:
Fixes https://github.com/magic-wormhole/magic-wormhole/issues/612

Some prototyping and documentation of explicit subprotocols per subchannel (for Dilation).

This changes:

- [x] no more need for "single-use-only" ``IStreamServerEndpoint`` from ``.dilate()``
- [x] ~~obvious how to implement a composable, re-usable Dilation sub-protocol~~
- [x] introduce a specialization of Twisted's ``IProtocolFactory`` to specify things
- [x] the control-channel speaks the empty subprotocol (``""``)
- [x] ~~subprotocol support (i.e. which ones you can use) are sent in to ``wormhole.create()`` e.g. instead of ``"_enable_dilation"``~~
- [x] define listening / connecting endpoints, with subprotocol names

Not all of the above is "done" in this branch yet, testing the concept out.

Please see the rendered documentation: https://magic-wormhole--617.org.readthedocs.build/en/617/

PR #623, opened 2025-05-26
Subprotos

shapr:
two typo fixes


meejah:
I merged this on top of `protype-subprotocols`

meejah:
shapr's commit from here is now in https://github.com/magic-wormhole/magic-wormhole/pull/617

PR #629, opened 2025-05-29
we dropped 3.9; update metadata

### LINKED ISSUE #592: Drop Python <= 3.9 support?
I used a couple union types in the (proposed) "feedback" / status branch. These are not critical, but are nice -- however, they require Python 3.10 or higher.

What concrete reasons are there to maintain Python 3.9 compatibility?

Regardless of what we decide here, Python 3.9 support ends in October this year (2025-10).

meejah:
fixes #592

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,6 +28,7 @@ setup(name="magic-wormhole",
       license="MIT",
       url="https://github.com/warner/magic-wormhole",
       classifiers=trove_classifiers,
+      python_requires=">=3.10",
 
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -29,6 +28,7 @@
       license="MIT",
       url="https://github.com/warner/magic-wormhole",
       classifiers=trove_classifiers,
+      python_requires=">=3.10",
```
hmaarrfk, line 31: nice.

PR #636, opened 2025-06-05
remove 'u' prefix to strings

sblondon:
It's unneeded and more consistent with the other strings.

In the patch, there are several `instance_of(type(u"")`. They could be replaced by `str`. I can rewrite the commit to replace them if you want.

Please see the rendered documentation: https://magic-wormhole--636.org.readthedocs.build/en/636/


meejah (approved):
LGTM! Thanks :)

PR #637, opened 2025-06-06
refactoring: use classes directly instead of type() calls

sblondon:
This PR follows the discussion in PR #636.

I also replaced `type(b'')`.
I did not replaced `instance_of()` calls. They seem less obvious changes.


meejah (approved):
LGTM thanks :)

PR #639, opened 2025-06-07
test_receive_{send,receive}: disable pytest's output capturing

mjoerg:
These tests might otherwise fail, unless pytest is invoked with --capture=no.


meejah:
Can you describe the actual problem? I've run master with ``pytest -s`` and ``pytest`` and they both run all the tests fine?

meejah:
Hmm, that's a weird error to get.

Yes, you usually pass ``pytest`` a directory / file to run but ...

PR #642, opened 2025-06-16
refactoring: replace formatted strings by f-strings

sblondon:
The change has been generated by [flynt tool](https://github.com/ikamensh/flynt). Four assertions failed in unit tests after the changes so those ones have been dropped. The differences are `\` transformed to `\\` so it could a `flynt` bug. I plan to investigate about it but it's out of scope here.

I didn't check a real use.


meejah:
Sorry for the delay, thanks for the contributions!

PR #644, opened 2025-06-17
Fix typo: "unparseable" → "unparsable" in test_hints.py

leopardracer:
**Description:**  
Corrects the spelling of "unparseable" to "unparsable" in test assertions for consistency and accuracy. No functional changes.


meejah:
Both are valid spellings .. not sure if it's worth changing the output format for this (although the output isn't intended for machines, but still)

PR #652, opened 2025-06-30
Fix #651

### LINKED ISSUE #651: QR Code doesn't work on dark displays
If you're using a "dark-mode" terminal like Solarized Dark, the QR code display doesn't work with typical phone detectors (only tested GrapheneOS built-in QR Scanner).

Presumably it needs to draw an explicitly "white" / light background and dark QR dots to work properly?

kisst:
Fix: QR Code doesn't work on dark displays

### DIFF
--- a/src/wormhole/cli/cmd_send.py
+++ b/src/wormhole/cli/cmd_send.py
@@ -139,7 +139,11 @@ class Sender:
         if not args.zeromode and args.qr:
             qr = QRCode(border=1)
             qr.add_data(f"wormhole-transfer:{code}")
-            qr.print_ascii(out=args.stderr)
+            # Use TTY colors if available, otherwise use ASCII
+            if args.stderr.isatty():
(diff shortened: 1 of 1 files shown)


src/wormhole/cli/cmd_send.py
```diff
@@ -139,7 +139,11 @@ def _go(self, w):
         if not args.zeromode and args.qr:
             qr = QRCode(border=1)
             qr.add_data(f"wormhole-transfer:{code}")
-            qr.print_ascii(out=args.stderr)
+            # Use TTY colors if available, otherwise use ASCII
+            if args.stderr.isatty():
+                qr.print_ascii(out=args.stderr, tty=True, invert=False)
```
meejah, line 144: Could maybe be simplified to ``..., tty=args.stderr.isatty()`` if you like?

meejah (approved):
Code looks good to me (optional inline comment). Please add a note to NEWS.md with your username (following the pattern).

Thanks!

PR #656, opened 2025-07-06
Fix #650 bare pytest does not work

### LINKED ISSUE #650: bare "pytest" doesn't work
For some reason, running just ``pytest`` in the root fails two tests (but running the way the makefile does works).

``pyproject.toml`` should be adjusted with whatever options PyTest needs to run properly without needing special options.

Hugo-C:
Hello, the problem come from both `-s`  and the test path being required. So on master this works:
```shell
pytest src/wormhole/test/test_args.py::test_receive_send
pytest -k test_receive_send -s
```
but this doesn't:
```sh
pytest -k test_receive_send
```

So it seems something outside the `test` directory is interfering with `sys.stdout` when it's collected by pytest. But I am not able to tell what it is. 
Anyway, we want to restrict which directories pytest browses. And specifying `testpaths` solves the issue.
 
 I dropped `-s` as it is not required. If you find it nice to have the live stdout/stderr, then `--capture=tee-sys` could be used to [have the best of both worlds](https://docs.pytest.org/en/stable/how-to/capture-stdout-stderr.html#setting-capturing-methods-or-disabling-capturing).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ lint:
 	tox -e flake8less
 
 test:
-	python -m pytest -W ignore::DeprecationWarning -r w -v -s src/wormhole/test
+	python -m pytest
 
 completions:
 	bash -c '_WORMHOLE_COMPLETE=bash_source wormhole > wormhole_complete.bash'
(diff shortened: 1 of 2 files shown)


meejah:
Perfect, thanks .. yeah I explored the differences a little and I'm also not sure what's going on with `sys.stdout`. There's Click involved as well, so could be a few things interacting...

meejah (approved):

PR #658, opened 2025-07-14
style: update syntax to recent Python versions

sblondon:
The update has been done with:
`pyup-dirs --py39-plus --recursive .`

(`pyup-dirs` is a `pyupgrade` for directories.)

Then some unused imports have been removed (because the uses have been removed previously by `pyup-dirs` command).

Tox is green after the changes (tested with python 3.13). Perhaps a tool like `pyup-dirs` should be added in a precommit file?

Please see the rendered documentation: https://magic-wormhole--658.org.readthedocs.build/en/658/


meejah (approved):
:+1: thanks!

PR #675, opened 2025-10-16
Link to SPAKE2 instead of SPAKE2+

teohhanhui:
SPAKE2 is balanced / symmetric. SPAKE2+ is augmented / asymmetric. It seems like magic-wormhole uses a shared password.

Please see the rendered documentation: https://magic-wormhole--675.org.readthedocs.build/en/675/


meejah:
Yes, magic-wormhole uses the symmetric / non-augmented one; good catch thanks!

meejah:
Merging anyway, not sure what's up with build

PR #685, opened 2025-11-24
NAT hole punching

meejah:
From @warner 's PoC code, a few robustness fixups, tests, and "vendoring" the Twisted patch.

Requires the ``"your-address"`` client IP + port reflection addition to the Mailbox (see https://github.com/magic-wormhole/magic-wormhole-mailbox-server/pull/63 )


meejah:
I'm still not entirely sure why the tests _need_ the retry logic. I think retrying once (only) is good here (e.g. if the TCP port you want is occupied) but testing manually with a local server and two local clients works -- and that _should_ be exactly the same as what the tests are doing (but: clearly something is different).

Comment out the `d.addErrback(_failed)` on line 713 of `transit.py` to see the tests fail

mistressxalexis (approved):

PR #709, opened 2026-04-10
fix: show user-friendly message for 'crowded' server error

### LINKED ISSUE #708: Improve user-facing error message for "crowded" server errors
1. When the magic-wormhole server is busy and returns a "crowded" error, users see an unhelpful technical error message (trace) instead of a clear explanation of what happened and what they should do.

2. Signature

```bash
> wormhole receive 00-foo-bar
  ...
wormhole.errors.ServerError: crowded
ERROR: crowded
```
3. Expected behavior: Users should see a clear, actionable message explaining that the server is busy and suggesting they try again.

4. Proposed solution: Modify the close_error method in src/wormhole/_boss.py to provide a user-friendly message for the "crowded" error case: 

src/wormhole/_boss.py (line 368-375)

```python
@m.output()
def close_error(self, errmsg, orig):
    if errmsg == "crowded":
        user_msg = ("The server is too busy to handle this request right now. "
                   "Please wait a moment and try again.")
        self._result = ServerError(user_msg)
    else:
        self._result = ServerError(errmsg)
    self._T.close("errory")
```

5. Impact:

- Low risk: Only affects error message formatting, not core functionality
- High value: Significantly improves user experience during server congestion
- Maintains backward compatibility: All other server errors behave the same

mvanhorn:
When the relay server returns "crowded" (too many clients on a single mailbox), users currently see a raw traceback ending in `ERROR: crowded`. This replaces that with an actionable message explaining the situation.

The wording follows @meejah's guidance in #708: the error means too many peers connected to the mailbox, which could be a typo or an attack, and the right action is to try again with a brand-new code.

**Changes:**
- `_boss.py`: `close_error` now checks for `errmsg == "crowded"` and substitutes a human-readable message. All other server errors pass through unchanged.
- `test_machines.py`: Added `test_server_error_crowded` that verifies the raw "crowded" string is replaced with the user-facing message.

Fixes #708

### DIFF
--- a/src/wormhole/_boss.py
+++ b/src/wormhole/_boss.py
@@ -366,16 +366,7 @@ class Boss:
 
     @m.output()
     def close_error(self, errmsg, orig):
-        if errmsg == "crowded":
-            self._result = ServerError(
-                "Too many peers have attempted to connect to this"
-                " mailbox. Try again with your intended peer, making"
-                " sure you get a brand-new code and that it is"
-                " transcribed correctly. If this keeps happening, it"
-                " may be due to an attack."
-            )
-        else:
-            self._result = ServerError(errmsg)
+        self._result = ServerError(errmsg)
         self._T.close("errory")
 
     @m.output()
--- a/src/wormhole/_boss.py
+++ b/src/wormhole/_boss.py
@@ -366,7 +366,16 @@ class Boss:
 
     @m.output()
     def close_error(self, errmsg, orig):
-        self._result = ServerError(errmsg)
+        if errmsg == "crowded":
+            self._result = ServerError(
+                "Too many peers have attempted to connect to this"
+                " mailbox. Try again with your intended peer, making"
+                " sure you get a brand-new code and that it is"
+                " transcribed correctly. If this keeps happening, it"
+                " may be due to an attack."
+            )
+        else:
+            self._result = ServerError(errmsg)
         self._T.close("errory")
 
     @m.output()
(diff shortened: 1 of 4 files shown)


meejah:
Thanks!

I think it might be better to "translate" the error for humans here https://github.com/magic-wormhole/magic-wormhole/blob/master/src/wormhole/cli/cli.py#L144 (approximately).

Basically the above `try/except` block could grow an `except` to handle `ServerError`, and interpret "crowded" there.

I'd have to double-check this and some other implementations, but the "server error strings" (like `"crowded"`) are at least somewhat specified. That is, I _believe_ that the `"crowded"` string matches one of these: https://github.com/magic-wormhole/magic-wormhole/blob/master/docs/server-protocol.rst#mailboxes and may be used / interpreted by other code. (If that is indeed true, this would be better modeled in the code with a proper type, but ...)

Nice work including a test :) and thanks for the PR

mvanhorn:
Thanks @meejah - you're right, CLI layer is the better place for this. Addressed in b0216ba:

- Reverted `_boss.py` `close_error` back to its original one-liner. `ServerError` keeps carrying the raw protocol code so library consumers can still match on it.
- `cli/cli.py` `_dispatch_command` grows an `except ServerError` handler that interprets `"crowded"` for humans and falls through to `"server error: <code>"` for other codes, mirroring the shape of the existing `ServerConnectionError` handler.
- Dropped `test_server_error_crowded` from `test_machines.py` (it was asserting on the boss-level translation that no longer exists) and added two CLI tests in `test_cli.py`: one that asserts `"crowded"` is absent and `"Too many peers"` / `"brand-new code"` are present in stderr, and one that asserts other server error codes are passed through.

All 68 tests in `test_machines.py` still pass and the new CLI tests are green locally.

meejah (approved):
LGTM!

Issue #144, opened 2017-03-30
use TLS for the rendezvous connection


warner:
It'd be a good idea to use TLS for the client's connection to the Rendezvous server. It doesn't actually protect that much, which is why it hasn't been a priority for me before:

* an eavesdropper watching the client's connection can still tell that they're using magic-wormhole, because the packets are aimed at the (public) rendezvous server's IP address
* an eavesdropper watching the server can correlate the traffic in and out to determine that client at IP-1 is talking to the client at IP-2, even if the messages themselves are encrypted
* an eavesdropper who can listen to two likely communication partners at the same time will be able to tell that they're talking wormhole to each other, based on the timing correlation, without even looking at the server

You need Tor to protect against the first one, and even then I suspect the message pattern is distinctive enough that an eavesdropper could figure it out.

To protect against the second two, in addition to TLS (to hide the exact channel/nameplate in use) and plenty of cover traffic, we also need clients to not be online at the same time (doing something like Pond's randomly-timed server connections, and fixed-size messages). This will require a feature I'm working on called "journaled mode" to work at all.

But adding TLS is still the right thing to do. On the client side, this is pretty simple: when we parse the rendezvous server URL, make sure we can accept `wss:` in addition to just `ws:`, build a `tcp:` or `tls:` endpoint descriptor string to match, then pass that string into `endpoints.clientFromString` instead of always using `HostnameEndpoint`. (we do need to enhance the TorManager to handle this too, running TLS over Tor over TCP, which might require changes to txtorcon, I'm not sure).

On the server side, wormhole-server needs to use txsni or an `endpoints.SSL4ServerEndpoint` instead of always using `TCP4ServerEndpoint`.

To get a valid certificate for the public rendezvous server, I plan to use LetsEncrypt. I had originally thought this would require changes to wormhole-server (to let it serve a static HTML directory, to publish the ACME protocol challenge/response messages). But someone pointed out to me that TLS certificates are not bound to any specific port number (the theory being that if you can control port 80, you probably own the entire box anyways).

So I just need to run a normal webserver on the rendezvous server host, listening on port 80, and configure it to participate in the LetsEncrypt protocol for the public rendezvous URL domain (relay.magic-wormhole.io). That instance of certbot needs to write the certificate and private key into a place where wormhole-server can see it. Then wormhole-server (running on port 4000) can use the certificate without actually knowing anything about LetsEncrypt.

The order of operations needs to be:

[...]

PR #715, opened 2026-05-13
public_relay.py: use TLS

dongdigua:
#144


meejah:
yes, it appears the main relay is now deployed behind a TLS terminator.

PR #716, opened 2026-05-13
docs(ecosystem.rst): add wormhole.page

dongdigua:
Please see the rendered documentation: https://magic-wormhole--716.org.readthedocs.build/en/716/


dongdigua:
https://github.com/lucamartinetti/wormhole-page seems to be entirely vibe coded btw

meejah (approved):

meejah:
Thanks. Yeah lots of "vibe" ;)

Still, it's "an app in the ecosystem" so I merged your PR, thanks!

I did look enough to confirm it's using the real default mailbox server, and works with the Python CLI. I believe it's doing transit via its own server AND the default transit-relay, which is a bit weird (e.g. if it's going to transit via their server anyway, just running their own transit relay would be better, since the latest transit-relay supports TCP/WS interop anyway).

PR #717, opened 2026-05-18
zizmore recommendations

meejah:


warner (approved):
Huge fan. Let's add comments next to these hashes to remind us that they were pinned from eg `actions/checkout@v4`, else it'll be harder to tell when we need to upgrade things. Feel free to land after adding those.

PR #728, opened 2026-08-20
bail with CrowdedError if we see messages from a third side

warner:
This changes the mailbox state machine to "imprint" upon the first side it sees, and reject any others by terminating with a CrowdedError. This is like WrongPasswordError (which indicates a cryptographic error) but merely means more than the two peers tried to use the same nameplate, usually because someone mistakenly attempted to reuse an old code.

A correctly-behaving mailbox server will not allow this to happen: when the third side appears, the server marks the mailbox as CROWDED, and clients who OPEN that mailbox (or CLAIM its nameplate) get back a "crowded" error before they can receive any of the extra messages. But a misbehaving server might let those messages get through, so clients should not accept them. And since a crowded mailbox means we don't really know which side is the "right" one, the safest thing to do is to terminate the wormhole and allow the user to try again.

### DIFF
--- a/src/wormhole/errors.py
+++ b/src/wormhole/errors.py
@@ -43,6 +43,8 @@ class LonelyError(WormholeError):
     """wormhole.close() was called before the peer connection could be
     established"""
 
+class CrowdedError(WormholeError):
+    """We received messages from multiple peers, there can be only one."""
 
 class WrongPasswordError(WormholeError):
     """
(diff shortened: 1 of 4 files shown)


src/wormhole/errors.py
```diff
@@ -43,6 +43,8 @@ class LonelyError(WormholeError):
     """wormhole.close() was called before the peer connection could be
     established"""
 
+class CrowdedError(WormholeError):
```
meejah, line 46: this probably needs _some_ kind of *Highlander* reference ;)

meejah (approved):
LGTM

PR #730, opened 2026-08-24
change CLI to use on_status_update= instead of w.get_unverified_key()

warner:
I'm looking to remove the w.get_unverified_key() API, because future versions of the PAKE protocol will use a different key for the key-confirmation message, so the final key won't be available early enough to provide anything like an "unverified" key. Also, it wasn't a good idea to expose the raw key anyways.

This changes the "wormhole send" and "wormhole receive" tools to manage their "waiting.." timers with the the on_status_update= callback API instead of get_unverified_key().

refs #732

### DIFF
--- a/src/wormhole/cli/cmd_receive.py
+++ b/src/wormhole/cli/cmd_receive.py
@@ -12,6 +12,7 @@ from twisted.python import log
 from wormhole import __version__, create, input_with_completion
 
 from ..errors import TransferError
+from .._status import WormholeStatus, AllegedSharedKey, ConfirmedKey
 from ..transit import TransitReceiver
 from ..util import (bytes_to_dict, bytes_to_hexstr, dict_to_bytes,
                     estimate_free_space)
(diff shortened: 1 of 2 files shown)


src/wormhole/cli/cmd_receive.py
```diff
@@ -118,6 +122,44 @@ def _bad(f):
...
+        if not self._slow_verification_timer:
+            self._slow_verification_timer = self._reactor.callLater(VERIFY_TIMER, self._on_slow_verification)
+
+    def _cancel_on_slow_verification_timer(self):
+        t = self._slow_verification_timer
+        if t and not t.called:
+            t.cancel()
+            self._slow_verification_timer = None
+
+    def _on_status(self, status):
```
meejah, line 151: I tried to push a commit suggesting a simpler way to use the status API (assuming we don't care about EOL'd python 3.9 anymore). I guess we'll see how that shows up in github ;)

meejah (approved):

src/wormhole/cli/cmd_receive.py
warner: thanks, I somehow missed `match` appearing in python, that's really nice

oh, but, the `if status.peer_key != self._old_status.peer_key` guard was meant to avoid restarting the timer on every status update, even the ones that don't change `.peer_key`. I'll update it to keep that guard but use `match` inside it

PR #733, opened 2026-08-24
remove get_unverified_key() and friends, update tests and docs

warner:
wormhole.get_unverified_key() and
delegate.wormhole_got_unverified_key() were removed. If the delegate
has such a method, emit a log.err warning at creation time. The docs
and Interface definitions were updated to match.

Tests were updated to not use get_unverified_key(). This required
tweaks to cmd_send.py/cmd_receive.py (to sample KEY_TIMER earlier) and
the conftest.py fixtures (to properly capture log.err() even when it
is called with a string instead of an Exception).

Please see the rendered documentation: https://magic-wormhole--733.org.readthedocs.build/en/733/

### DIFF
--- a/src/wormhole/wormhole.py
+++ b/src/wormhole/wormhole.py
@@ -2,7 +2,7 @@ import os
 import sys
 
 from attr import attrib, attrs
-from twisted.python import failure
+from twisted.python import failure, log
 from twisted.internet.task import Cooperator
 from zope.interface import implementer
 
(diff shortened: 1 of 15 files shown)


meejah:
sorry, i pushed a thing as "a suggestion" before noting the 3 stacked PRs, so I've done a couple pushes (and force-pushes, blame my "jj" brain) to add some release-notes for the next release. Feel free to change those, just intended as a suggestion :)

src/wormhole/wormhole.py
```diff
@@ -283,6 +276,8 @@ def create(
     cooperator = Cooperator(scheduler=eq.eventually)
 
     if delegate:
+        if hasattr(delegate, "wormhole_got_unverified_key"):
+            log.err(AttributeError("WARNING: wormhole_got_unverified_key has been removed, see #732"))
```
meejah: Maybe `DeprecationWarning` should be used here?

meejah (approved):
LGTM, see one inline comment

PR #737, opened 2026-08-27
refactor key setup, prepare for new protocol versions

warner:
This rewrites the key-setup and encryption core of the magic-wormhole protocol into a sans-io -style EncryptionCore and a collection (one for now) of pluggable INegotiation classes, in preparation for versions of the key-setup protocol.

This "Negotiate_V0" protocol should behave indistinguishably from the old code, but it's in a better spot to work in parallel with future "v1"/etc protocols.


meejah:
A few notes after looking at factoring this like either old-style or new-style Automat machine (partially to match the rest, partially "just to see").

(I haven't gone through the latest PR in detail yet though, just wanted to put these down)

- the "get_outputs()" dance seems extremely similar to the "self-input" stuff that new-Automat fixes
- new-automat forces you to use a `typing.Protocol` for inputs, which is kind of the same purpose INegotiation serves
- I think 'side' and 'phase' belong "outside" (i.e. something else checks sides, and the 'phase' should  be burned into the input symbols, maybe?)
- "who deals with order" is somewhat ambiguous 

The "input" -> "process()" -> "process_pake()" dance is like the "conditional state" thing in new-Automat. That said, I still think it might be a lot cleaner to do the parsing etc outside, and feed in symbols like `NegotiationOutput` does -- that is, a "PakeZero" and a "Versions" message for now (growing more pake-phase messaging later). This would put a lot more of the "parsing-adjacent" stuff outside and might read better? (e.g. checking 'phase' strings for validity feels like parsing, checking 'side' values feels like parsing or sanity-checks).

-----

**mocks / tests**

Some of this is definitely well outside the scope of _this_ PR, but I find the "mock"-based tests very hard to follow and read (and to convince myself we're testing "API" / observable things vs. internal state). It would be very nice to have a way to test at least these new pieces in more isolation and without mocks. Asserting on events like "M_AddMessage" feels a lot like this whole flow is there "for testing" / tracing (and the event-list replaces mocks)?

I'm hoping to frame this as some concrete changes to the tests -- my intuition says that since this is p2p we should just be able to "cross-wire" two negotiation instances and see them come to a successful conclusion. At least in the  happy-path cases. For example, in `test_good_key` instead of hard-wiring the "other peer" into the tests via asserts on the events, it seems we should be able to create a second instance (and thus avoid a second copy of the SPAKE2 protocol in the tests).

meejah:
Okay, I hacked together a new-Automat style state-machine.

It ends up conflating "order" into the states. I pushed the "handle side / crowded" into the higher layer (`_EncryptionCore`). This is what that machine ends up looking like (from automated diagrams out of Automat). I'll push some code in a minute, but note that the "handle crowded" is currently a "TODO" in `_EncryptionCore` ;)

This machine could be simplified further if `_EncryptionCore` also handled ordering. Then one might argue that we haven't "gotten rid" of the `Order` machine at all but merely made it "not an explicit state-machine".

There are probably more things that should lead to "error", but all the existing unit-tests pass. I also changed the API a little so that all "outputs" are just returned from "input" calls to the state-machine, so there's one less "queue-like abstraction". Only the I/O layer can know when it's appropriate to have to "do a reactor loop" or similar, so that handing really needs to be in the I/O layer anyway.

It's "nearly" a drop-in for the non-explicit-state-machine processor (but the "outputs" API switch needs a couple other lines swapped around too).

wormhole _key_setup key_setup_v0 key_setup_factory dot

Both styles still need cleanup on "when do we deliver the key" (i.e. only after `"version"` arrives).

Edit: PoC code is here https://github.com/meejah/magic-wormhole/pull/new/736-refactor-key-setup--automat

warner:
I rebased slightly, should be the same overall diff as before but 1: based on current trunk and 2: merged the last two commits so that the branch refactors directly into `KeySetup_V0` instead of briefly using the wrong name.

I plan to replace the "`thing.output()` returns one event" with "each method returns a list of events", like you said, before this will be ready for review.

Re wiring-two-instances-together.. I agree that is easier, but it's not really confirming that the code behaves in a way that matches the spec. If the code misbehaves but does so in a way that's compatible with it's own buggy self, then other implementations that *do* follow the spec won't interoperate.

The main thing I want from tests like that is to check that all the possible permutations of inputs are covered. The expected outputs will vary depending on the sequence of inputs, for example if Negotiation learns the peer's `my_key_setup_versions` list before it is asked to create the initial `phase="pake"` message (what I'm calling the PAKE-0 in the new negotiation code), then it doesn't have to speculate about the possible versions, which lets it do less work and make the message smaller. The overall test flow is like:

* invoke input API X
* check that it outputs messages 1,2,3
* invoke input API Y
* check that it outputs zero messages
* invoke input API Z
* check that it outputs message 4
* etc

One caveat is that it may have to pretend to be the peer, so e.g. `Y` may depend on output message `2`, so there's some code in the middle of the test that's copied out of the implementation. We could use another instance of the `KeySetup_V0` or whatever to do that work, but it's not designed to emit pieces in useful-for-the-test order, and if e.g. the implementation were accidentally changed in some spec-incompatible way, such a test wouldn't catch it.

I'm trying to wrangle the tests into a shape that are easier to follow. Lemme pull some of the later improvements into this branch and see if they look more appealing.

meejah:
I  think the best path for "exploratory" tests are going to be Hypothesis plus their "State-based" testing, where you teach a state-machine what the entire state-space is and let it explore. This is more for like "does the implementation blow up in the face of weird messages".

Regardless, having the infrastructure to do either of these things would ideally make it pretty straightforward to write tests "by hand" when we _do_ encounter 'real-world, non-Python implementations' that cause problems.

In some sense, the mock-based tests are just another implementation. I don't know enough TLA+ to actually-use that here right now, but that would be another avenue to explore (and probably the best bet as far as "yes, this follows the spec").

As I said elsewhere, I'll take a stab at "new-style Automat-ing" these. I have written one very simple state-based test so could also see how that goes at the same time.

src/wormhole/_encryption.py
```diff
@@ -0,0 +1,244 @@
...
+__all__ = ["Encryption", "_EncryptionCore"]
+# phase classifiers
+
+DILATE_RE = re.compile(r'^dilate-(\d+)$')
+NUMERIC_RE = re.compile(r'^\d+$')
+
+def is_key_setup(phase):
+    return phase == "pake" or phase.startswith("pake-") or phase == "version"
+def is_dilation(phase):
+    return bool(DILATE_RE.search(phase))
```
meejah: Probably want match() not search here?

src/wormhole/_key_setup/spake2_helper.py
```diff
@@ -0,0 +1,24 @@
+from spake2 import SPAKE2_Symmetric
+from ..util import to_bytes
+
+class SPAKE2_Helper:
```
meejah: What is the purpose of this?

(It _looks_ like it is so we can call start() and/or finish() any number of times and get the same output?)

src/wormhole/_key_setup/spake2_helper.py
```diff
@@ -0,0 +1,24 @@
...
+        assert code == self._code
+        if not self._msg1:
+            code_b = to_bytes(code)
+            appid_b = to_bytes(self._appid)
+            self._sp = SPAKE2_Symmetric(code_b, idSymmetric=appid_b)
+            self._msg1 = self._sp.start()
+        return self._msg1
+    def finish(self, msg2):
+        assert not self._finished
+        self._finished
```
meejah: `self._finished = True` presumably? (as-is this line doesn't do anything)

src/wormhole/_key_setup/key_setup_v0.py
```diff
@@ -0,0 +1,108 @@
...
+            self._error = self._error or CrowdedError()
+        if self._error:
+            raise self._error
+        actions = False
+        next_wanted = False
+        match self._state:
+            case Init():
+                raise ValueError("input() before start")
+            case WantPAKE(wanted):
+                assert phase == wanted
```
meejah: I feel like errors might be better handled as a `Error(message)` output from this method. Same for the Crowded or PasswordError. As-is, this is a separate path for "error-adjacent" messages.

Regardless, this `assert` should be a proper exception like the others, or a nice error-message to send when we shut down the protocol.

src/wormhole/_key_setup/negotiator.py
```diff
@@ -0,0 +1,116 @@
+from attrs import frozen, define, field
+from zope.interface import implementer
+from . import inegotiator, ikeysetup
+from .key_setup_v0 import KeySetup_V0
+from .._interfaces import ITiming
+from ..util import dict_to_bytes, provides
+
+# constructors are sampled at construction time, for unit tests
```
meejah: I'll play with some alternatives, but this seems like an awkward API.

It looks like this all exists so that `Negotiator` (via `_build_negotiator`) can instantiate the right thing. It would be more clear to pass in the right thing, or pass in "a function that can build the right thing" at least.

Even though this exists "for unit-tests" I find that it's extremely often the case that a "tests only" API kind of _wants_ to be public, or at least indicates that the existing API isn't flexible enough. This stuff is pretty "internal" but I think this metric still applies.

src/wormhole/_key_setup/negotiator.py
```diff
@@ -0,0 +1,116 @@
...
+    def got_code(self, code: str) -> None:
+        match self._state:
+            case Waiting():
+                key_setup = self._build_negotiator()
+                self._state = self._start_one_version(key_setup, code)
+                self._drain_inbound()
+            case Negotiating() | Done():
+                raise IllegalCall
+            case _:
+                raise UnknownState
```
meejah: this seems truly like an "internal error" / "really really should never happen" sort of thing that doesn't need a custom exception?

src/wormhole/test/common.py
```diff
@@ -157,3 +158,23 @@ async def poll_until(predicate):
         d = defer.Deferred()
         reactor.callLater(0.001, d.callback, None)
         await d
+
+
+class Dummy:
+    def __init__(self, name, events, iface, *meths, **kw):
```
meejah: `**kw` is never actually used.

This could be more clear if the two things that ever use this `IMailbox` and `IBoss` just had the relevant message declared on their interfaces, and this `Dummy` could be like `RecordingStateMachine` or similar and inspect the incoming interface to fake out everything on it.

Also `self.retval` is never used either.

src/wormhole/_encryption.py
```diff
@@ -0,0 +1,250 @@
...
+    def _drain_queued_sends(self):
+        assert self._key
+        while self._queued_sends:
+            (phase, plaintext) = self._queued_sends.pop(0)
+            data_key = derive_phase_key(self._key, self._side, phase)
+            encrypted = encrypt_data(data_key, plaintext)
+            self._add_output(M_AddMessage(phase, encrypted))
+
+
+# and this class is the IO-capable frontend
```
meejah: note to selves: this should also mean it's the only thing that can or should care about time, or "eventual sends" or anything like that.

src/wormhole/_encryption.py
```diff
@@ -0,0 +1,250 @@
...
+    def wire(self, boss, mailbox):
+        self._B = IBoss(boss)
+        self._M = IMailbox(mailbox)
+
+    def set_trace(self, _tracer):
+        pass # unimplemented on non-Automat machines for now
+
+    def _process_actions(self, actions):
+        # ideally these would be eventual-sends, but I don't want to
+        # build that much runtime, so we just use self._events as a
```
meejah: I've seen this done this in a pretty straightforward way in other repositories: you pass in an "eventual-send" function to `__init__` which uses `reactor.callLater(0, ...)` in real code and just immediately calls the function in test-code.

I don't personally _love_ it but it seems like a good compromise (if your goal is "having tests that don't use a proper reactor").

src/wormhole/_encryption.py
```diff
@@ -0,0 +1,250 @@
...
+        self._process_actions(actions)
+
+    # input from Boss and Dilation
+    def send(self, phase, plaintext):
+        actions = self._core.send(phase, plaintext)
+        self._process_actions(actions)
+
+    # shortcut methods for unit tests
+
+    def test_count_received_messages(self):
```
meejah: As you may have gathered, my instincts are to regard "just for testing" APIs as a code-smell.

This appears to be used as an awaiting tactic (only) in tests. Since this class _is_ allowed to "do I/O" it can thus have an `async` method -- and this seems like a perfect candidate. It could also then grow the ability to wait for certain events (as hinted at in the testing comments -- later versions might need a more-expressive thing than just "wait for one message to arrive").

Hopefully I'll have a better suggestion after a deeper dive.

src/wormhole/util.py
```diff
@@ -3,10 +3,17 @@
...
+from hashlib import sha256
+
 from cryptography.hazmat.primitives.kdf import hkdf
 from cryptography.hazmat.primitives import hashes
+from nacl.secret import SecretBox
+from nacl import utils as nacl_utils
+from nacl.exceptions import CryptoError
 from attr import attrs, attrib
 
+CryptoError # re-export
```
meejah: Might be safer to wrap this?

That is, have our own `class CryptographyError` with our own ("safe for users") message, possibly also with the wrapped / real NaCl exception kept around for logging purposes.

Otherwise, we're tying our public API to strict versions of NaCl and might have issues if they upgrade their exceptions in a way that we deem incompatible to our users. Something along the lines of:

```
@define
class CryptographyError(Exception):
    nacl_error: CryptoError
    message: str
```

src/wormhole/util.py
```diff
@@ -117,3 +124,41 @@ def provides(interface):
         value it got.
     """
     return _ProvidesValidator(interface)
+
+
+def derive_key(key, purpose, length=SecretBox.KEY_SIZE):
+    if not isinstance(key, bytes):
```
meejah: p.s. there is some kind of runtime type checking we can opt into. I don't know all the details, and well beyond the scope of this PR, but might be worth a tracking ticket -- that is, if we want to put typing information into magic-wormhole, "runtime checks" could be one use of it. I think :)

meejah (commented):
Some preliminary comments after a read-through on GitHub.

I'll play with the branch some more (including Automat-ing, as discussed) and probably have some better comments /  opinions after that ;)

